PR #1, opened 2020-11-04
Update package.xml, update ur_description launching

AndyZe:
Before this, the packages did not show up with `ros2 pkg list`

I also snuck in a URDF update. This launch file works now:

`ros2 launch ur_description view_ur5.launch.py`   --> Launches RViz but the robot isn't displayed yet

`ros2 topic echo /tf`  (to verify it's working)

@zultron


zultron:
Looking good.  This builds and runs in my environment.

PR #2, opened 2020-11-05
Packaging updates and CI configuration

zultron:
This PR further adjusts packaging so that `rosdep` can compute dependencies.

It adds a `moveit_ci` configuration that at least partly passes on Travis CI; the main tests fail because the source-based dependencies run over the max 45 minute time limit on Travis.  See [this build][1] for example.

[1]: https://travis-ci.com/github/zultron/Universal_Robots_ROS2_Driver/builds/198565397


nbbrooks:
Do you have an idea of which dep (hopefully just one) is using up the bulk of the time?

zultron:
> Do you have an idea of which dep (hopefully just one) is using up the bulk of the time?

These are the biggies:
```
Finished <<< moveit_core [20min 36s]
Finished <<< moveit_planners_ompl [15min 5s]
Finished <<< moveit_servo [11min 47s]
Finished <<< moveit_ros_robot_interaction [6min 23s]
```

zultron:
...although this may actually succeed now, since caching is in place.  I broke something since that previous, promising CI run; let me fix it and maybe we'll see this pass.

Universal_Robots_Client_Library/CMakeModules/DefineCXX17CompilerFlag.cmake
```diff
@@ -1,27 +1,29 @@
...
 #
 #   <var>  - variable to store the resulting flag
 #
-# Use this to find the compiler option to enable C++17 compilation. This is usefull
-# on cmake versions / use cases which do not support CXX_STANDARD.
-# Use this in conjunction with target_compiler_option():
+# Use this to find the compiler option to enable C++17
+# compilation. This is usefull on cmake versions / use cases which do
+# not support CXX_STANDARD.  Use this in conjunction with
+# target_compiler_option():
```
AndyZe, line 17: I'm just curious, did clang format trim the text like this?

Universal_Robots_Client_Library/package.xml
```diff
@@ -1,14 +1,9 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>ur_client_library</name>
   <version>0.1.1</version>
```
AndyZe, line 5: Perhaps we should set the version to 0.0.0 for now

deprecated_ros1_pkgs/controller_stopper/package.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
```
AndyZe, line 3: For these packages in /deprecated_ros1_pkgs, I wasn't planning to build them. They don't seem necessary in the short term.

Just a note so you don't spend time on this deprecated folder in the future.  (Maybe we should just delete the folder?)

ur_description/package.xml
```diff
@@ -17,11 +15,16 @@
...
+  <author>Kelsey Hawkins</author>
+  <author>Mathias Ludtke</author>
+  <author>Felix Messmer</author>
+
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <exec_depend>joint_state_publisher</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
-  <exec_depend>rviz</exec_depend>
+  <exec_depend>rviz2</exec_depend>
```
AndyZe, line 27: good catch!

.repos.yaml
```diff
@@ -0,0 +1,45 @@
+repositories:
```
AndyZe, line 1: It would be great if you added instructions on using this to the README. Is it something like this:

`vcs import < Universal_Robots_ROS2_Driver/.repos.yaml`

AndyZe: I think this is a good way to do it

Universal_Robots_Client_Library/CMakeModules/DefineCXX17CompilerFlag.cmake
AndyZe: It's strange that it's trimming around 70-character length. Usually it's 80 or 120, I think. Should be 120 to match MoveIt2, I believe

AndyZe: Possibly should just revert the changes to this file

AndyZe (commented):
Make the addition to the README, set the version to 0.0.0, and address my text formatting question, then it looks good to merge!

If you agree, I think we can merge it without CI passing. Since we're trying to move fast

gavanderhoorn:
Just curious: why are we building MoveIt2 from sources in the CI setup of a robot driver?

AndyZe:
That's a good point.

Another issue is that MoveIt2 froze the version of ros2_control they're using at an older commit (https://github.com/ros-planning/moveit2/blob/a810037e697c25e8df74a049fed6cca55ff8e874/moveit2.repos#L23).

I guess moveit2 should be removed from the build for now.

Universal_Robots_Client_Library/CMakeModules/DefineCXX17CompilerFlag.cmake
zultron: > I'm just curious, did clang format trim the text like this?

The `ament_cmake_lint` test complained (all errors are pasted into the commit log):

    Universal_Robots_Client_Library/CMakeModules/DefineCXX17CompilerFlag.cmake:4: Lines should be <= 80 characters long [linelength]

Emacs `M-q` did the actual wrapping; you're right, the `fill-column` variable default is `70`.  I added a `.dir-locals.el` file to set that variable to `80`, matching `ament_cmake_lint`.

Universal_Robots_Client_Library/package.xml
zultron: Done.  Thanks!

deprecated_ros1_pkgs/controller_stopper/package.xml
zultron: Ha ha, with my emacs macros, it's actually more work to make exceptions!

The `ur_robot_driver/launch/ur_control.launch` file still refers to `controller_stopper`.  I'll just remove `ur_calibration` for now.

.repos.yaml
zultron: Done.

zultron:
> Just curious: why are we building MoveIt2 from sources in the CI setup of a robot driver?

Hello, Gijs!

This package uses [moveit_ci][1], which runs `rosdep install -y -q -n --from-paths . --ignore-src --rosdistro foxy`.  That runs through all `package.xml` files searching for dependencies' `rosdep` keys.  There are no such keys for packages in the `moveit2` distribution, since that hasn't been released to Foxy yet.  So instead I added the `.repos.yaml` file to `$UPSTREAM_WORKSPACE` so `moveit_ci` can satisfy the `rosdep`s by building from source.

I don't see any way to add a `rosdep install --skip-keys [moveit-related rosdep keys]` argument in `moveit_ci`.  If there's a better way, I'm all ears!  This is a pretty heavy build, and before `ccache` is primed, it exceeds the Travis CI job time limit.

[1]: https://github.com/ros-planning/moveit_ci

zultron:
Tests are failing on compiler warnings, but otherwise the linters and regression tests are passing.

The main/`g++` build warnings all resemble the following,  `-Wredundant-decls` warnings from the `ur_dashboard_msgs` package in `.cpp` code generated from `SetMode.action` by the `rosidl` `fastrtps` packages.

```
/opt/ws_moveit/build/ur_dashboard_msgs/rosidl_typesupport_fastrtps_cpp/ur_dashboard_msgs/action/detail/dds_fastrtps/set_mode__type_support.cpp:684:6: warning: redundant redeclaration of ‘bool ur_dashboard_msgs::action::typesupport_fastrtps_cpp::cdr_serialize(const SetMode_Goal&, eprosima::fastcdr::Cdr&)’ in same scope [-Wredundant-decls]
  684 | bool cdr_serialize(
      |      ^~~~~~~~~~~~~
/opt/ws_moveit/build/ur_dashboard_msgs/rosidl_typesupport_fastrtps_cpp/ur_dashboard_msgs/action/detail/dds_fastrtps/set_mode__type_support.cpp:31:1: note: previous declaration of ‘bool ur_dashboard_msgs::action::typesupport_fastrtps_cpp::cdr_serialize(const SetMode_Goal&, eprosima::fastcdr::Cdr&)’
   31 | cdr_serialize(
      | ^~~~~~~~~~~~~
```

The `clang` build fails on warnings from the `ur_client_library` package, all `-Wunused-parameter` warnings except for one `-Wdelete-abstract-non-virtual-dtor` warning.

```
In file included from /opt/ws_moveit/src/Universal_Robots_ROS2_Driver/Universal_Robots_Client_Library/examples/rtde_client.cpp:28:
In file included from /opt/ws_moveit/src/Universal_Robots_ROS2_Driver/Universal_Robots_Client_Library/include/ur_client_library/rtde/rtde_client.h:31:
[1m/opt/ws_moveit/src/Universal_Robots_ROS2_Driver/Universal_Robots_Client_Library/include/ur_client_library/comm/pipeline.h:215:36: [0m[0;1;35mwarning: [0m[1munused parameter 'name' [-Wunused-parameter][0m
  virtual void started(std::string name)
[0;1;32m                                   ^
```

Unless watchers have an easy fix, I'm not going to try to fix these for this PR.  All other review comments are addressed.  Ready for merge!

gavanderhoorn:
> This package uses [moveit_ci](https://github.com/ros-planning/moveit_ci), which runs `rosdep install -y -q -n --from-paths . --ignore-src --rosdistro foxy`. That runs through all `package.xml` files searching for dependencies' `rosdep` keys. There are no such keys for packages in the `moveit2` distribution, since that hasn't been released to Foxy yet. So instead I added the `.repos.yaml` file to `$UPSTREAM_WORKSPACE` so `moveit_ci` can satisfy the `rosdep`s by building from source.
> 
> I don't see any way to add a `rosdep install --skip-keys [moveit-related rosdep keys]` argument in `moveit_ci`. If there's a better way, I'm all ears! This is a pretty heavy build, and before `ccache` is primed, it exceeds the Travis CI job time limit.

Which package is bringing in MoveIt as a dependency?

Afaik, none of the packages in `Universal_Robots_ROS_Driver` depend on it.

AndyZe:
I think I'll get this PR merged because @zultron needs to move on to other things. In a follow-up PR, I'll try to address @gavanderhoorn's comments, which are:

- ignore ur5_moveit_config in the CI build

- Do not include moveit2 in .repos.yaml

(if I understand correctly)

AndyZe:
Also, use a fork of Universal_Robots_Client_Library instead of copy/pasting into this repo

PR #4, opened 2020-11-09
Delete deprecated pkgs: ur_bringup & controller_stopper

AndyZe:
I don't believe **ur_bringup** has been used since the original **ur_driver** package. And I don't believe **controller_stopper** is necessary for the ROS2 port (but can be added later)

@gavanderhoorn @fmauch (if you have time)


fmauch:
I agree with both points.

gavanderhoorn:
`ur_bringup` has never been used with `ur_robot_driver`. Why did you copy it @AndyZe ?

AndyZe:
I just wasn't aware that it wasn't used anymore. There are a lot of UR packages to keep track of

PR #10, opened 2020-11-27
Add universal robots e-series decription and configuration files

livanov93:
Prepare UR e-series robots packages for  ROS2


gavanderhoorn:
Could I suggest to port the packages in this branch: [ros-industrial/universal_robot@melodic-devel-staging](https://github.com/ros-industrial/universal_robot/tree/melodic-devel-staging).

Let's not perpetuate the `ur_e_*` packages, except for the MoveIt packages.

gavanderhoorn:
I'm also not entirely sure I'd put the MoveIt configurations in the same repository, tbh.

livanov93:
I agree about having one ur_description package. 
Regarding moveit packages, for first trajectory demonstration I will have these within the driver repository. They can and will be deleted afterwards, to make the package look more like [ROS1 version](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver). This should follow after universal_robots package migration towards `ament_cmake`.

AndyZe:
> Regarding moveit packages, for first trajectory demonstration I will have these within the driver repository. They can and will be deleted afterwards

Let's not be so certain yet. There are a lot of groups involved and there will probably be some discussion about it

livanov93:
> Let's not be so certain yet. There are a lot of groups involved and there will probably be some discussion about it

Then we won't rush with these  package organization decisions.
Moving the robot on ROS2 is the top priority.

AndyZe:
It probably would have been best to start from the `develop` branch, since what you're adding here doesn't depend on the hardware_interface PR (as far as I can tell).

livanov93:
I've set it to develop as it should be. You're right, this doesn't depend on the [HW PR](https://github.com/PickNikRobotics/Universal_Robots_ROS2_Driver/pull/8).

livanov93:
Added in #14

PR #32, opened 2021-01-20
Use only position command interface for Beta driver

livanov93:
This PR removes all velocity command interface stuff on actively developed branch as a result of milestones for Beta driver release.

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -393,7 +393,8 @@ return_type URPositionHardwareInterface::write()
       if (pos_diff_sum != 0.0)
         position_interface_in_use_ = true;
     }
-    else if (position_interface_in_use_)
+
+    if (position_interface_in_use_)
     {
       ur_driver_->writeJointCommand(urcl_position_commands_, urcl::comm::ControlMode::MODE_SERVOJ);
     }
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -150,6 +150,8 @@ return_type URPositionHardwareInterface::start()
 {
   RCLCPP_INFO(rclcpp::get_logger("URPositionHardwareInterface"), "Starting ...please wait...");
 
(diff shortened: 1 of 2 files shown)


ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -374,35 +376,26 @@ return_type URPositionHardwareInterface::write()
...
-
-    // create a lambda substract functor
-    std::function<double(double, double)> substractor = [](double a, double b) { return std::abs(a - b); };
-
-    // create a position difference vector
-    std::vector<double> pos_diff;
-    pos_diff.resize(urcl_position_commands_.size());
-    std::transform(urcl_position_commands_.begin(), urcl_position_commands_.end(), urcl_position_commands_old_.begin(),
-                   pos_diff.begin(), substractor);
+    if (!position_interface_in_use_)
```
destogl, line 379: With this logic here you will not call `writeJointCommand` in the same update call when you detect `position_interface_in_use_ = true;`

Is this correct? The old velocity implementation will do this in the same loop.

destogl (changes requested):
Did you intend to remove velocity control completely? Why not control this completely using `positio_interface_in_use_` flag?

The code is clean and everything is fine, but I have too little information about motivation to do this like it is done.

ur_robot_driver/src/hardware_interface.cpp
livanov93: Correct!
https://github.com/PickNikRobotics/Universal_Robots_ROS2_Driver/pull/32/commits/4aad52397470f02486bd9bc269fbb368361dad00

livanov93:
> Did you intend to remove velocity control completely? Why not control this completely using `positio_interface_in_use_` flag?
> 
> The code is clean and everything is fine, but I have too little information about motivation to do this like it is done.

Yes, decided to fully remove it for Beta driver milestone. Also to avoid any misunderstandings for the driver users for now if they take a look at the code (commenst like there is a method writeJointCommand in velocity mode, but the driver does not support it...).

destogl (approved):

Issue #36, opened 2021-02-04
Test small movements in joint space


nbbrooks:
Test tolerance failing of trajectory completion with small movements

MGBla:
Ok just tested it out on a single joint (wrist_1). Up till 0.1 radians the action terminates successfully. Everything above result in the following message:
``` 
[ros2_control_node-1] [INFO] [1612449853.661850170] [ur_joint_trajectory_controller]: Received new action goal
[ros2_control_node-1] [INFO] [1612449853.661883916] [ur_joint_trajectory_controller]: Accepted new action goal
[ros2_control_node-1] [WARN] [1612449863.664272377] [ur_joint_trajectory_controller]: Aborted due to goal tolerance violation
``` 
For reproducing the error, I started the robot at home pose [0.0, -1.57, 0.0, -1.57, 0.0, 0.0] and send the following ActionGoal

``` 
control_msgs.action.FollowJointTrajectory_Goal(trajectory=trajectory_msgs.msg.JointTrajectory(
header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), 
joint_names=['elbow_joint', 'shoulder_lift_joint', 'shoulder_pan_joint', 'wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint'], 
points=[trajectory_msgs.msg.JointTrajectoryPoint(positions=[0.0, -1.57, 0.0, -1.3, 0.0, 0.0], 
velocities=[], 
accelerations=[], 
effort=[], 
time_from_start=builtin_interfaces.msg.Duration(sec=10, nanosec=0))]), 
path_tolerance=[control_msgs.msg.JointTolerance(name='elbow_joint', position=3.14, velocity=3.14, acceleration=3.14), control_msgs.msg.JointTolerance(name='shoulder_lift_joint', position=3.14, velocity=3.14, acceleration=3.14), control_msgs.msg.JointTolerance(name='shoulder_pan_joint', position=3.14, velocity=3.14, acceleration=3.14), control_msgs.msg.JointTolerance(name='wrist_1_joint', position=3.14, velocity=3.14, acceleration=3.14), control_msgs.msg.JointTolerance(name='wrist_2_joint', position=3.14, velocity=3.14, acceleration=3.14), control_msgs.msg.JointTolerance(name='wrist_3_joint', position=3.14, velocity=3.14, acceleration=3.14)], 
[...]

PR #37, opened 2021-02-07
Fix joint_trajectory_controller execution - set stopped_velocity_tolerance to zero

livanov93:
@MGBla Addresses #36


MGBla:
It seems like the parameters are not added to the parameter service. There are only the following parameters:
``` 
/ur_joint_trajectory_controller:
  action_monitor_rate
  allow_partial_joints_goal
  constraints.goal_time
  constraints.stopped_velocity_tolerance
  joints
  state_publish_rate
  use_sim_time
```
As stated in #36 the root of the error was the stopped_velocity_tolerance.
But on a general note, I really don't think we should specify the tolerances in this yaml file. Since the JointTrajector msg specifies all the tolerance in its definition I think we would circumvent the intended mechanics here.

PR #43, opened 2021-02-18
Scaled Joint Trajectory Controller

MGBla:
Implementation of the ScaledJointTrajectoryController based on the JointTrajectoryController of ros2 control


MGBla:
@livanov93 Branch is ready for review. 
There is one discussion point for the goal_time_tolerance when the controller scales down the movement, which I will address in another issue.

MGBla:
The UR arms have an internal speed scaling which slows down all joints if for example the maximum joint speed configured in the teach panel is exceeded. This also includes the speed slider on the bottom of the teach panel.
So when the speed of all joints is scaled down the basic JTC action fails due the violation of the path tolerance.
Basically we want the same behavior as the JTC but first have to gather the current speed scaling factor L53-60(.cpp file)
The idea (also used in the ROS1 UR driver) is to use the scaling factor to calculate a scaled trajectory time L120-126 and use this timestamp in the sampling of the trajectory instead of node_->now() L130-131.
Since this is more a property of UR arms and not generally applicable I'm not sure this should be handled in the general JTC.

MGBla:
@destogl But just to clarify I wouldn't mind if the update loop could be parametrized with a time stamp or something like this in your rewrite. This way we could scale down on the copy paste of the update function in the derived controller which would make the code more robust, if further parts of the JTC change in the future.

ur_controllers/src/scaled_joint_trajectory_controller.cpp
```diff
@@ -19,10 +19,212 @@
...
+
+controller_interface::return_type ScaledJointTrajectoryController::update()
+{
+  if (state_interfaces_.back().get_name() == "speed_scaling")
+  {
+    scaling_factor_ = state_interfaces_.back().get_value();
+  }
+  else
+  {
+    RCLCPP_ERROR(get_node()->get_logger(), "Speed scaling interface not fount. This should not have happended");
```
livanov93, line 59: found

ur_controllers/src/scaled_joint_trajectory_controller.cpp
```diff
@@ -19,10 +19,212 @@
...
+  JointTrajectoryPoint state_current, state_desired, state_error;
+  const auto joint_num = joint_names_.size();
+  resize_joint_trajectory_point(state_current, joint_num);
+
+  // current state update
+  for (auto index = 0ul; index < joint_num; ++index)
+  {
+    state_current.positions[index] = joint_position_state_interface_[index].get().get_value();
+    state_current.velocities[index] = joint_velocity_state_interface_[index].get().get_value();
+    state_current.accelerations[index] = 0.0;
```
livanov93, line 104: Hardcoded value is ok?

livanov93 (changes requested):
Only small comments. Take a look and change. It does not request additional review. Will merge immediately after spelling is changed in the logger output.

ur_controllers/src/scaled_joint_trajectory_controller.cpp
MGBla: This was copied from the JTC. I think the acceleration is currently not set. Was the same in ROS1 if I remember correctly

livanov93 (approved):

PR #55, opened 2021-03-07
Use GitHub Actions instead of Travis

livanov93:


livanov93:
This PR should be ignored!

Issue #26, opened 2021-01-08
dashboard client node


nbbrooks:

livanov93:
#5 already addressed porting `dashboard_client_node` to ROS2. All the service servers have to be tested on the real hardware.

MGBla:
+ /add to log -> works
+ /connect -> [dashboard_client-2] terminate called after throwing an instance of 'rclcpp::exceptions::ParameterAlreadyDeclaredException'                                                                                                                   
[dashboard_client-2]   what():  parameter 'receive_timeout' has already been declared                                                                                                                                                         
[ERROR] [dashboard_client-2]: process has died [pid 19490, exit code -6,
+ /get_loaded_program -> works
+ /get_robot_mode -> works
+ /get_safety_mode -> works
+ /load_installation -> Not sure yet
+ /load_program -> Not sure yet
+ /popup -> works
+ /program_running -> works
+ /program_saved ->works
+ /program_state ->works
+ /quit -> seems to disconnect the robot but could not verify if all states are correct since connect throws errors
+ /raw_request -> not sure yet what it should do
+ /topic -> has shut down the robot not sure if intended behavior shouldn't quit do this?

MGBla:
Pull request ready for review

PR #60, opened 2021-03-17
Dashboard client fixes

MGBla:
Addresses #26 
Changes
* Fixed issue during capsuled trigger service creation
* Fixed issue when connecting mutliple times
* Added try catch blocks since it is no longer done during the service call in ROS2
* Start dashboard client from within hardware interface
* Made use of dashboard client namespace for service calls

### DIFF
--- a/ur_robot_driver/src/dashboard_client_ros.cpp
+++ b/ur_robot_driver/src/dashboard_client_ros.cpp
@@ -32,6 +32,7 @@ namespace ur_robot_driver
 DashboardClientROS::DashboardClientROS(const rclcpp::Node::SharedPtr& node, const std::string& robot_ip)
   : node_(node), client_(robot_ip)
 {
+  node_->declare_parameter<double>("receive_timeout", 1);
   connect();
 
   // Service to release the brakes. If the robot is currently powered off, it will get powered on on the fly.
@@ -199,7 +200,9 @@ bool DashboardClientROS::connect()
 {
   timeval tv;
   // Timeout after which a call to the dashboard server will be considered failure if no answer has been received.
-  tv.tv_sec = node_->declare_parameter<double>("receive_timeout", 1);
+  double time_buffer;
+  node_->get_parameter("receive_timeout", time_buffer);
+  tv.tv_sec = time_buffer;
   tv.tv_usec = 0;
   client_.setReceiveTimeout(tv);
   return client_.connect();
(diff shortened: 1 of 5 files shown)


ur_robot_driver/src/dashboard_client_ros.cpp
```diff
@@ -224,15 +317,24 @@ bool DashboardClientROS::handleRunningQuery(const ur_dashboard_msgs::srv::IsProg
 bool DashboardClientROS::handleSavedQuery(ur_dashboard_msgs::srv::IsProgramSaved::Request::SharedPtr req,
                                           ur_dashboard_msgs::srv::IsProgramSaved::Response::SharedPtr resp)
 {
-  resp->answer = this->client_.sendAndReceive("isProgramSaved\n");
-  std::regex expected("(true|false) ([^\\s]+)");
-  std::smatch match;
-  resp->success = std::regex_match(resp->answer, match, expected);
+  try
```
destogl, line 320: Maybe these almost the same block should be put in an external function in the future.  Just an idea.

ur_robot_driver/src/dashboard_client_ros.cpp
```diff
@@ -241,111 +343,129 @@ bool DashboardClientROS::handleSavedQuery(ur_dashboard_msgs::srv::IsProgramSaved
...
+      }
+      else if (match[1] == "VIOLATION")
+      {
+        resp->safety_mode.mode = ur_dashboard_msgs::msg::SafetyMode::VIOLATION;
+      }
+      else if (match[1] == "FAULT")
+      {
+        resp->safety_mode.mode = ur_dashboard_msgs::msg::SafetyMode::FAULT;
+      }
+      // The following are only available in SafetyStatus from 5.5 on
```
destogl, line 390: Should this be parameterized by the firmware version? Would this be feasible in the future. If so, could you open a follow-up issue.

destogl (approved):
Nice and clean! Thanks! Just few comment to open follow-up issues if you think it something to consider in the future. 

@MGBla you are free to merge...

livanov93 (approved):
I am all good with this one!

ur_robot_driver/src/dashboard_client_ros.cpp
MGBla: Since the msg response differs between the different calls this would not work. For all Trigger Services this was already done

MGBla:
One of you two has to merge it, since I'm not allowed to do it

PR #61, opened 2021-03-28
Delete the UR5 moveit_config package

AndyZe:
My understanding is, we want to focus on the E-series (UR5e, UR3e, etc). So, delete this package.


urrsk:
It should work on both CD3 and E-series. Just like the ROS driver and Client Library

AndyZe:
It's just that the scope of the initial work was E-Series, and we already have a lot on our plate to ship that soon.

Issue #50, opened 2021-03-02
Fix motion back to home position bug


nbbrooks:

nbbrooks:
@MGBla it is my understanding you are tackling this issue, is that correct? If not, please unassign youself - thanks!

destogl:
https://github.com/ros-controls/ros2_controllers/pull/140

PR #63, opened 2021-04-08
Move to home bug

MGBla:
Addresses #50


MGBla:
This is just a quick fix for preventing unwanted behavior of the real robot at startup. The real fix will come in a later pull request.

destogl (approved):
Looks good and if it works, great!

PR #81, opened 2021-04-13
Added view_ur for checking description

destogl:
As per title.


destogl:
I'll just merge this because the changes are simple and it does not influence other files.

PR #82, opened 2021-04-14
Make an optional launch arg for RViz, document it in README

AndyZe:
This adds an argument for RViz launching and updates the README about it.

Also, I changed some numbers in the README to bullet points since they don't really need to be traversed in order anymore.


destogl:
Great idea! I only moved the argument to the end of the arguments list, to start with UR-related arguments.

destogl (approved):

PR #83, opened 2021-04-14
Merge ROS2 beta driver development into main

destogl:
Format fix-up is in #84.

Please merge this without squashing keeping the history.


AndyZe (approved):
I gave this a final test. It looks good! I will let CI finish (I don't expect all jobs to pass just yet) then merge it.

PR #88, opened 2021-04-19
Setting speed slider with range of 0.0-1.0 and added warnings if range is exceeded

MGBla:


destogl (approved):
It looks good. I just updated formatting.

PR #92, opened 2021-04-19
Fix move to home bug

MGBla:
* Before sending data to the robot the hardware interface not verifies if the controller has send new data or if the current joint values are just stale output.
* Fixed a timing issue in the scaled trajectory controller. When executing multiple trajectories subsequently, the scaled timing was not handled correctly which caused a delay of the starting point of the trajectory.
* Removed **Attention** Flags related to move to home bug in the Readme file


destogl (approved):
This looks very good. Thanks!

Issue #90, opened 2021-04-19
IO Controller: Protective Stop after setting slider to 0


MGBla:
After setting the Speed Slider to 100 using the io controller service, the robot goes into protective stop giving the following warning:  
![ur](https://user-images.githubusercontent.com/35597356/115217733-1474b280-a106-11eb-9772-d4ee617da17c.jpg)

MGBla:
The issue seems a problem in the client library and is addressed in https://github.com/UniversalRobots/Universal_Robots_Client_Library/issues/58

MGBla:
Circumvented the issue for the moment by not allowing the value to be set below 1%.

PR #100, opened 2021-05-03
Fixed speed slider service call

### LINKED ISSUE #89: IO Controller: Setting speed slider to 1.0
Currently the speed slider cannot be set to 100%. The controller seems to accept the input but on the call to the hardware interface something seems to go wrong. Thus the request is ignored

MGBla:
Fixes #89 and #90

### DIFF
--- a/ur_controllers/src/gpio_controller.cpp
+++ b/ur_controllers/src/gpio_controller.cpp
@@ -266,16 +266,15 @@ bool GPIOController::setSpeedSlider(ur_msgs::srv::SetSpeedSliderFraction::Reques
   if (req->speed_slider_fraction >= 0.01 && req->speed_slider_fraction <= 1.0) {
     RCLCPP_INFO(node_->get_logger(), "Setting speed slider to %.2f%%.", req->speed_slider_fraction * 100.0);
     // reset success flag
-    command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_ASYNC_SUCCESS].set_value(2.0);
+    command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_ASYNC_SUCCESS].set_value(ASYNC_WAITING);
     // set commanding value for speed slider
     command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_CMD].set_value(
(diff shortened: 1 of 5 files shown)


ur_controllers/src/gpio_controller.cpp
```diff
@@ -263,22 +263,23 @@ bool GPIOController::setIO(ur_msgs::srv::SetIO::Request::SharedPtr req, ur_msgs:
...
     // reset success flag
-    command_interfaces_[CommandInterfaces::SCALING_ASYNC_SUCCESS].set_value(2.0);
+    command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_ASYNC_SUCCESS].set_value(2.0);
     // set commanding value for speed slider
-    command_interfaces_[CommandInterfaces::SPEED_SCALING_CMD].set_value(
+    command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_CMD].set_value(
         static_cast<double>(req->speed_slider_fraction));
 
-    while (command_interfaces_[CommandInterfaces::SCALING_ASYNC_SUCCESS].get_value() != 2.0) {
+    while (command_interfaces_[CommandInterfaces::TARGET_SPEED_FRACTION_ASYNC_SUCCESS].get_value() != 2.0) {
```
urrsk, line 274: Please define use constants for numbers that have a specific meaning.

urrsk (changes requested):
2.0 and -1 seems to be constants with a specific meaning. Please define them as constants with a meaningful name, and use that instead. It will also make the code more readable.

ur_controllers/src/gpio_controller.cpp
MGBla: Changed it so that the shared cmd variable contains a NaN if no new data is available and capsuled it into a static constant

MGBla: Ok still have to have a look at what @livanov93 intended with the async communication at this point.

MGBla: Exchanged 2.0 value as well and removed a race condition between controller and hardware interface in the async check

urrsk (approved):
Looks good

Issue #65, opened 2021-04-08
Use Client Library of the Universal Robots Github


MGBla:
Currently we are working on a fork of the Client Library repo with additional ROS2 changes. However, since the Client Library is plain c++ it should support both the ROS1 and ROS2 driver without additional branches.

bblumberg:
This is additional information on #93. Unfortunately, I don't have the log file anymore. But, I can provide more details. The official UR client library did build (despite catkin warnings) and I was able to link against it. However, at runtime, using a real ur3e, the controller-manager (?) crashed when I used the sample launch instructions from the documentation. I remember it was a malloc error which is what led me to think that you might have had your own version of the client library as you did. When I set the fake hardware option to true, it worked.Sorry I didn't save the log file.

MGBla:
Ok thats weird, I have kind of the exact same setup with an ur3e here. I will try to reproduce the error later on a freshly checked out  workspace

bblumberg:
Hey, I rebuilt things using the UR version of the UR client library, and
got the same error as yesterday, but this time i saved the launch.log file.
Hope it helps.

bb

MGBla:
As it seems there is currently an issue in the ur_client_library which sound rather similar. There seems to be a problem if the ros-noetic-ur-client-library is installed while simultaneously having the client library in a workspace. Do you have the package installed from a privious use of the ROS1 driver? The current workaround is to remove the apt package but the client library driver team is already working on this.

bblumberg:
It seems like a different problem. I am installing in a fresh ros:foxy
container and here is how I am installing  the client library
git clone
https://github.com/UniversalRobots/Universal_Robots_Client_Library.git
urcl-ur && \
cd urcl-ur && \
mkdir build && cd build && \
cmake .. && \
make && \
make install

For what it is worth, when I use the PickNik version of the client library,
I can load controllers, but when I try the joint_trajectory controller, it
sends out commands, but the robot doesn't move. I am getting joint states
etc. so I know that the communication at some level is working.

bb

MGBla:
Ok maybe it is the std build process which is failing at some level. Normally i clone the client repo into the colcon workspace and just use ``` colcon build ```  to build the client library and the ros2 driver simultaneously, without installing it into the system.

bblumberg:
Thanks, I'll give that a try.

bb

bblumberg:
[...]

PR #101, opened 2021-05-03
Using official Universal Robot Client Library

MGBla:
This should solve #65

### DIFF
--- a/Universal_Robots_ROS2_Driver.repos
+++ b/Universal_Robots_ROS2_Driver.repos
@@ -15,10 +15,10 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_controllers.git
     version: master
-  ur_client_library:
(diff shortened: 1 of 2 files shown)


urrsk:
Any reason why not use the urscript in the client library?

MGBla:
There is still a problem with the client library and ROS2 which i couldn't solve yet. The lib works and compiles but since it is a cmake package it is not added to the ament_index. As a result we are not able to find the client library in any launch files where the urscript is specified. I found this pose( https://answers.ros.org/question/341870/ros2-pkg-list-show-cmake-packages/) which is roughly related but was not yet able to get the client library correctly registered on the ament_index. Therefore as a workaround for the moment I just updated the urscript to the most recent version in the driver itself where we can find it from launch files.

urrsk (approved):
Looks good

PR #124, opened 2021-06-01
Converted io_test and switch_on_test to ROS2

urmahp:
The tests can be run with colcon and with launch_test

Headless_mode is not implemented in ROS2 therefore trajectory_test cannot be converted to ROS2 yet.


ur_robot_driver/test/integration_test.py
```diff
@@ -0,0 +1,180 @@
...
+
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "ur_type", default_value="ur5e", description="Type/series of used UR robot."
+        )
+    )
+
+    ur_type = LaunchConfiguration("ur_type")
+    robot_ip = LaunchConfiguration("robot_ip")
+    dir_path = os.path.dirname(os.path.realpath(__file__))
```
destogl, line 57: Would it make sense to use here `ThisLaunchFileDir`?

Include: `from launch.substitutions import ThisLaunchFileDir`
Use: `ThisLaunchFileDir()` - see e.g. [here](https://github.com/ros-controls/ros2_control_demos/blob/4888549d2f8ad6de9e025bdf62ad4541b6674729/ros2_control_demo_bringup/launch/rrbot_system_position_only.launch.py#L69)

ur_robot_driver/test/integration_test.py
```diff
@@ -0,0 +1,180 @@
...
+
+    launch_file = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource([dir_path, "/../../ur_bringup/launch/ur_control.launch.py"]),
```
destogl, line 60: Actually, with the previous comment, this line can be written as:

```suggestion
        PythonLaunchDescriptionSource([ThisLaunchFileDir(), "/../../ur_bringup/launch/ur_control.launch.py"]),
```

destogl (approved):
Small optimization of the launch part of the test file. Otherwise good. We have to also repair CI.

ur_robot_driver/test/integration_test.py
urmahp: It doesn't seem that ThisLaunchFileDir() can find the current directory, when used inside a test file. I think it has to be used inside a launch file.

urmahp:
> Small optimization of the launch part of the test file. Otherwise good. We have to also repair CI.

I am pretty sure it should be enough to just update [the branch](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/.github/workflows/ci-ursim.yml#L6) to master and develop in the URSim CI, I will be happy to do it here or in a different PR?

destogl:
@urmahp I just rebased this to main. Let's see if the tests are passing.

urmahp:
The test seems to be executed inside the coverage build and also inside the binary build, so we should probably have the simulator running on those pipelines as well. 

There are also some packages, that the test cannot locate when executed in the pipeline, which makes the pipeline fail. I haven't seen this locally, but I will have a look at it first thing tomorrow.

destogl:
This looks good. Thanks!

PR #125, opened 2021-06-04
Use ur_client_library from released version

fmauch:
As the client library has been released to ROS foxy, it might be good to
also use this inside CI. This would also help when making changes to the client
library, as the downstream build shadows the current PR when being installed from
source instead of using the version from the current build workspace.

### DIFF
--- a/Universal_Robots_ROS2_Driver.repos
+++ b/Universal_Robots_ROS2_Driver.repos
@@ -15,10 +15,6 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_controllers.git
     version: 0.2.1
-  Universal_Robots_Client_Library:
-    type: git
-    url: https://github.com/UniversalRobots/Universal_Robots_Client_Library.git
-    version: master
   ur_msgs:
     type: git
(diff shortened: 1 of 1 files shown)


fmauch:
The binary build seems to fail as the released version uses the `boost` branch of the library instead of the `master` branch. Therefore, it doesn't export a c++17 dependency and therefore the build fails on all c++17 functionalities from this driver.

I decided to release the foxy version also from the boost branch, as the [target platform specification}(https://www.ros.org/reps/rep-2000.html#foxy-fitzroy-may-2020-may-2023) specifies c++14 for Foxy as minimum requirement.

It might be either a good idea to add the c++17 requirement to this driver or release the library's foxy package from the `master` branch. Any opinions?

destogl:
Probably we should reconsider here if C++14 can be used... I am not sure which functionalities from C++17 are used at all. Opening a follow-up issue.

destogl (approved):

destogl:
Replaced by #142 where binary build uses binary and semi-binary uses build from source.

PR #141, opened 2021-07-10
Correct obsolete ros2controlcli flags

destogl:
the `--state` flag became obsolete and `--set-state` should be used.


livanov93 (approved):
LGTM!

Issue #140, opened 2021-07-08
could not determine a constructor for the tag '!degrees'


NebuDa:
Hi,

I have tried to install the ROS2 Driver with this Howto: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver

At the Driver Testing I got the following Error:

```
flo@DESKTOP-L4I9CIG:~$ ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=true
[INFO] [launch]: All log files can be found below /home/flo/.ros/log/2021-07-08-13-15-59-881791-DESKTOP-L4I9CIG-138
[INFO] [launch]: Default logging verbosity is set to INFO
Task exception was never retrieved
[...]

Issue #131, opened 2021-06-20
Build failing with ros2_controllers installed from apt


sanatmharolkar:
Hi, 

I already had the ros2_control packages installed from apt using the command "sudo apt install ros-foxy-ros2-control ros-foxy-ros2-controllers ros-foxy-control-msgs". After cloning the UR driver packages, I made sure the dependencies were satisfied using rosdep. So then I tried to build the UR driver packages in my colcon workspace. However the build fails with the following message:

```
/home/sanat/colcon_ws/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp: In member function ‘virtual controller_interface::return_type ur_controllers::ScaledJointTrajectoryController::update()’:
/home/sanat/colcon_ws/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:58:10: error: ‘is_halted’ was not declared in this scope; did you mean ‘is_halted_’?
   58 |     if (!is_halted) {
      |          ^~~~~~~~~
      |          is_halted_
/home/sanat/colcon_ws/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:59:7: error: ‘halt’ was not declared in this scope
   59 |       halt();
      |       ^~~~
/home/sanat/colcon_ws/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:93:38: error: ‘joint_position_state_interface_’ was not declared in this scope; did you mean ‘joint_state_interface_’?
   93 |     state_current.positions[index] = joint_position_state_interface_[index].get().get_value();
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                      joint_state_interface_
/home/sanat/colcon_ws/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:94:39: error: ‘joint_velocity_state_interface_’ was not declared in this scope; did you mean ‘has_velocity_state_interface_’?
   94 |     state_current.velocities[index] = joint_velocity_state_interface_[index].get().get_value();
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                       has_velocity_state_interface_
[...]

PR #142, opened 2021-07-10
Update CI configration to supprot galactic and rolling

destogl:
Changes:

﻿- Make binary build really a build based on released binaries only. This should help detect issues like #140 and #131
- Add binary builds for galactic and rolling.
- Base coverage, build on rolling.
- Add source builds for all distros and correct the branch names there.
- Removes dedicated ursim build because it is integrated into testing of all other builds.


destogl:
@livanov93 and @nbbrooks Now all checks are passing. We can build for `foxy`, `galactic` and `rolling`. I also added ccache in builds. Please review and merge if you are happy.

livanov93 (approved):
Looks good to me! Thanks!

livanov93 (approved):
#162 will continue after merging this

PR #149, opened 2021-08-19
Fixes access to the "prefix" property

jrgnicho:
Minor change to avoid the need to creat a "prefix" xacro:arg in calling xacros

### DIFF
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -82,7 +82,7 @@
     <xacro:include filename="$(find ur_description)/urdf/ur.ros2_control.xacro" />
     <!-- ros2 control instance -->
     <xacro:ur_ros2_control
-      name="URPositionHardwareInterface" prefix="$(arg prefix)"
+      name="URPositionHardwareInterface" prefix="${prefix}"
       use_fake_hardware="$(arg use_fake_hardware)"
       fake_sensor_commands="$(arg fake_sensor_commands)"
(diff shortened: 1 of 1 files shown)


ur_description/urdf/ur_macro.xacro
```diff
@@ -82,7 +82,7 @@
     <xacro:include filename="$(find ur_description)/urdf/ur.ros2_control.xacro" />
     <!-- ros2 control instance -->
     <xacro:ur_ros2_control
-      name="URPositionHardwareInterface" prefix="$(arg prefix)"
```
destogl, line 85: Isn't this access to an internal variable and not an attribute?

Would it be a solution to add default value to attribute in L57? what do you think?

jrgnicho: The $(arg var) syntax does access an internal argument defined with **xacro:arg** if it exists, otherwise it'll look for any arguments with the same name defined in the calling  xacro.  I think it is preferable to set this as a parameter to the macro as that allows more control over the naming of the entities; this is especially useful if you have multiple robots in a workcell and one needs to apply different prefixes for each arm.

destogl: Oh I see, this is wrong you are right. Thans for the fix!

destogl (approved):

jrgnicho:
@destogl thanks for approving, would you be able to merge?

destogl (approved):

nbbrooks (approved):

PR #152, opened 2021-08-30
Modifiy parameter declaration - approach equalization with ros-controls dependencies

livanov93:
PR info:

- adapts `ur_controllers` package which components inherit from [`ControllerInterface`](https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/controller_interface.hpp)
- follows changes in [ros2_control](https://github.com/ros-controls/ros2_control/pull/504/files) and [ros2_controllers](https://github.com/ros-controls/ros2_controllers/pull/224)

### DIFF
--- a/ur_controllers/src/force_torque_sensor_broadcaster.cpp
+++ b/ur_controllers/src/force_torque_sensor_broadcaster.cpp
@@ -33,10 +33,10 @@ controller_interface::return_type ForceTorqueStateBroadcaster::init(const std::s
 
   try {
     auto node = get_node();
-    node->declare_parameter<std::vector<std::string>>("state_interface_names", {});
-    node->declare_parameter<std::string>("sensor_name", "");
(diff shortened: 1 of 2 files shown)


ur_controllers/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -33,10 +33,10 @@ controller_interface::return_type ForceTorqueStateBroadcaster::init(const std::s
 
   try {
     auto node = get_node();
```
destogl, line 35: Did you forget this line?

```suggestion
```

destogl (approved):
Small comment, otherwise good. the same changes as in ros-controls/ros2_controllers#224

ur_controllers/src/force_torque_sensor_broadcaster.cpp
livanov93: yep

PR #154, opened 2021-09-01
Allows setting initial joint positions for Generic system

jrgnicho:
This PR adds the initial_position argument to allow passing a dictionary with initial joint positions when using the Simulated Controller.  This dictionary can also be loaded from a yaml  file as follows:
```<xacro:property name="initial_positions" value="${load_yaml(initial_positions_file)}"/>```  where the initial_positions_file looks as follows:
```
elbow_joint: 1.158
shoulder_lift_joint: -0.953
shoulder_pan_joint: 1.906
wrist_1_joint: -1.912
wrist_2_joint: -1.765
wrist_3_joint: 0.0
```


destogl:
Can you explain the reasons for this change. IMO these changes add complexity which is not adding much simplicity on the other side. These changes are probably needed only once or very rare so having them as parameters in yaml file does not add to simplicity.

urrsk:
@destogl it might be useful in the CI setting up the screen!
@jrgnicho can you share your motivation for the PR?

jrgnicho:
This allow passing an initial position to the robot so that it doesn't start with all the joints at zero.  I have a workcell where the robot is in collision with an object at startup because the robot is fully extended at the zero joints position

henningkayser:
I think this change makes a lot of sense. It's debatable if we want to default to using a yaml file as input or an alternative way where the initial values are passed to the node directly (so that we don't have to feed them into the xacro). To me this is a very common use case and we should find a best practice that allows overriding initial positions without having to copy the xacro file.

jrgnicho:
@livanov93 I made the requested change

destogl:
@livanov93 just to not forget. We should also port this to the main branch after merging

rcywongaa:
Note this currently does not work with galactic in simulation
Pending PR: https://github.com/ros-controls/ros2_control/pull/635
Related: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/305

Issue #165, opened 2021-09-14
 xacro using __limits["..."] is invalid


livanov93:
With ros/xacro@9941961 names starting with __ are considered illegal.
It leads to crashing on startup.

PR #166, opened 2021-09-14
Fix invalid xacro properties

livanov93:
Addresses #165


destogl (approved):
Looks good. There seems to be a missing rolling sync on ros2_control-side.

PR #178, opened 2021-09-22
Retrun foxy CI config to main to work with sheduled builds.

destogl:
Scheduled workflows work only on main/default branches:
https://stackoverflow.com/questions/63436541/github-action-workflow-schedule-not-working-on-non-default-branch/63437012#63437012


livanov93 (approved):

urrsk:
Maybe we can use repository-dispatch to trigger a schedule build from the main branch to other branches?
https://github.com/peter-evans/repository-dispatch

PR #185, opened 2021-09-26
Rename the joint controller that is started by default

AndyZe:
If I understand right, `robot_controller` is the joint controller that is launched by default. I think that's a little confusing because there are several controllers running on the robot. Would prefer to name it `initial_joint_controller` to specify exactly what it does.

Currently, the options in ur_controllers.yaml are `joint_trajectory_controller` and `forward_velocity_controller`


destogl (approved):
Thanks! This adds clarity for sure!

destogl:
The rolling test is currently expected to fail until the next release.

livanov93:
This PR should have changed README instructions

Issue #168, opened 2021-09-16
Parameter naming in joint_limits.yaml


christianlandgraf:
Hi,

when trying to test the Pilz Planner in ROS 2 with the UR, I noticed `joint_limits.yaml` in the ROS 2 driver specifies the limits with regard to a generic name (e.g. wrist_3, wrist_2 etc) instead of the actual joint name (e.g. wrist_3_joint, wrist_2_joint) in the urdf.

See [ROS 1](https://github.com/fmauch/universal_robot/blob/calibration_devel/ur10e_moveit_config/config/joint_limits.yaml) and [ROS 2](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_description/config/ur10e/joint_limits.yaml)

Is there any reason for changing this?
Since the Pilz motion planner checks if velocity limits (and acceleration limits) exist for each joint in a planning group, it does not correctly load the parameters (e.g. [here](https://github.com/sjahr/moveit2/blob/pr-port_pilz_planner/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_ptp.cpp#L79)).

I didn't encounter a problem somewhere else in MoveIt, so far, but theoretically it could appear in other parts as well?

destogl:
This seems rather as an issue with wrong configuration of joint to plan for. You have to specify them somewhere with the suffix `_joint`. Please try to check this.

@livanov93, @urrsk do you think we should change names as they were in ROS1?

livanov93:
@destogl The thing is that links that have been compared are not from the same packages. The [ROS1](https://github.com/fmauch/universal_robot/blob/calibration_devel/ur10e_moveit_config/config/joint_limits.yaml) version that @christianlandgraf  has referenced is from `urXYZ_moveit_config` package and [the one ROS2 version referenced](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_description/config/ur10e/joint_limits.yaml) is in `ur_description` package.
If we take a look in [ROS1 version](https://github.com/fmauch/universal_robot/blob/calibration_devel/ur_description/config/ur5e/joint_limits.yaml) of `ur_description` it is obvious it has the same form (without `_joint` suffix).

[...]

PR #187, opened 2021-09-27
Joint limits parameters for Moveit planning

livanov93:
Closing #168


gavanderhoorn:
Would this be an acceptable solution?

It introduces duplicates of information and doesn't copy over the acceleration nor effort limits.

At least in the past, not having acceleration limits caused MoveIt to fall back to `1.0` rad/s^2, which results in rather 'tame' trajectories.

gavanderhoorn:
you're missing the point.

@AndyZe is advocating for including *jerk* limits in `joint_limits.yaml`, we have quite some information on UR limits, but then this PR goes in the opposite direction by ignoring all of that and then copying a limited subset.

livanov93:
I wasn't aware of ros-planning/moveit2#706, but thanks for pointing that out. However, @AndyZe will also have somehing to say on this topic.

ur_moveit_config/launch/test_moveit_servo_velocity_controller.launch.py
```diff
@@ -260,7 +271,10 @@ def generate_launch_description():
 
     kinematics_yaml = load_yaml("ur_moveit_config", "config/kinematics.yaml")
 
-    joint_limits_yaml = {"robot_description_planning": joint_limit_params}
+    joint_limits_yaml = load_yaml(
```
AndyZe, line 274: Same comment about not wanting to duplicate the loading of joint_limits.yaml here.

ur_bringup/launch/ur_moveit.launch.py
```diff
@@ -249,6 +260,11 @@ def generate_launch_description():
     kinematics_yaml = load_yaml("ur_moveit_config", "config/kinematics.yaml")
     robot_description_kinematics = {"robot_description_kinematics": kinematics_yaml}
 
+    joint_limits_yaml = load_yaml(
```
AndyZe, line 263: Similar to what gavanderhoorn said, I would prefer not to duplicate the joint_limits.yaml that was loaded at L160. Can you re-use that parameter?

ur_moveit_config/config/ur10/joint_limits.yaml
```diff
@@ -0,0 +1,34 @@
+# joint_limits.yaml allows the dynamics properties specified in the URDF to be overwritten or augmented as needed
+# Specific joint properties can be changed with the keys [max_position, min_position, max_velocity, max_acceleration]
+# Joint limits can be turned off with [has_velocity_limits, has_acceleration_limits]
```
AndyZe, line 3: I think we can add acceleration, effort, and jerk limits in a followup PR, eventually  :+1:

ur_bringup/launch/ur_moveit.launch.py
livanov93: I can, and will do this in alternative way. I'll ping you when it is finished

livanov93: 7c862d6

ur_moveit_config/launch/test_moveit_servo_velocity_controller.launch.py
livanov93: 7c862d6

ur_bringup/launch/ur_moveit.launch.py
```diff
@@ -20,6 +20,7 @@
 from ament_index_python.packages import get_package_share_directory
```
destogl, line 20: We should at some point test creating `xml`-launch files . This here are getting really nasty.

ur_moveit_config/launch/test_moveit_servo_velocity_controller.launch.py
```diff
@@ -24,6 +24,7 @@
 from ament_index_python.packages import get_package_share_directory
 from launch import LaunchDescription
 from launch.actions import DeclareLaunchArgument
+from launch.actions import OpaqueFunction
```
destogl, line 27: can we here somehow reuse at least part of the`ur_moveit.launch.py`-file? This huge launch files are getting un-maintanable.

destogl (commented):
I don't feel competent to comment on limits, but just two style-related comments.

ur_bringup/launch/ur_moveit.launch.py
livanov93: It is not that bad. Also I am not sure how many examples are there about xml launch files. We started to catch the grip with the python-based one :)

AndyZe:
I get this. Any idea what I'm missing?

`yaml.constructor.ConstructorError: could not determine a constructor for the tag '!degrees'`

AndyZe:
Ah, found this related PR you made previously:  https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/166

livanov93:
@AndyZe #140 was caused because of the problem in `xacro`. However, this is work in progress and to be able to parse joint_limits yaml the make it work with moveit components **! degrees** have to be changed to radians, or another yaml-parsing procedure has to be implemented as in [xacro](https://github.com/ros/xacro/blob/b8716fc1b72c13c8f9799ccd3a6d4f840dca4097/xacro/__init__.py#L129)

PR #188, opened 2021-09-27
Update Servo parameters, for smooth motion

AndyZe:
- update the publish_period. 50 Hz is too slow for UR robots, it seems.
- launch a Servo node rather than a pose tracker node. Pose tracking is a little more specialized and complicated, so I don't think it's needed for this demo launch file

https://user-images.githubusercontent.com/11284393/134968581-512d244d-a952-4c0a-869b-0916247f4ac3.mp4

### DIFF
--- a/ur_bringup/launch/ur_control.launch.py
+++ b/ur_bringup/launch/ur_control.launch.py
@@ -114,7 +114,7 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument(
             "robot_controller",
-            default_value="joint_trajectory_controller",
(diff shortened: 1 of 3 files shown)


livanov93:
@AndyZe Are joint limits working and used correctly by servo? Take a look at UniversalRobots/Universal_Robots_ROS2_Driver#187

AndyZe:
I didn't think to check joint limits. Sorry.

If it works with regular MoveIt motion planning, e.g. in RViz, it should be fine for Servo too, though.

livanov93 (approved):
Great to move from `pose_tracking` to `servo` node!

PR #191, opened 2021-09-30
Fix prepare and perform switch operation

### LINKED ISSUE #190: Prepare and perform switch operation fails
Sometimes, the `controller_manager` while switching (starting) multiple controllers (ft sensor ctrl, io_ctrl, ...) turns of the [flags](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/a939a570a3104befef1514def08c6a6dc82702b8/ur_robot_driver/include/ur_robot_driver/hardware_interface.hpp#L193-L194) that detect if position or velocity based controller is currently active. This results in `controller_manager` reporting active status for the commanding controller, but the hardware is not sending commands

livanov93:
PR solves the problem of reporting the controller status active while its belonging command interface is disabled in the system.

Closes #190

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -643,19 +643,22 @@ return_type URPositionHardwareInterface::perform_command_mode_switch(const std::
 {
   hardware_interface::return_type ret_val = hardware_interface::return_type::OK;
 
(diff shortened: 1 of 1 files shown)


destogl (approved):
Looks good, but can you give more information about testing scenario?

PR #224, opened 2021-10-26
Use gpio tag in urdf

destogl:
This is now fully functional, and it should be tested on hardware with  UniversalRobots/Universal_Robots_ROS2_Description#16.


fmauch (approved):
LGTM!

Tested with URSim and seems to work fine, as well.

PR #226, opened 2021-10-26
Add resource files from ROS World.

destogl:


ur_robot_driver/doc/resources/README.md
```diff
@@ -0,0 +1,30 @@
+# Resources about ROS 2 Universal Robot driver
+
+The files in this folder are available for use under [![License](https://i.creativecommons.org/l/by/4.0/88x31.png)](http://creativecommons.org/licenses/by/4.0/) license.
```
nbbrooks, line 3: This image looks a little awkward when inline with text. Can we just saw Creative Common license (and maybe move the image to below the text?)

ur_robot_driver/doc/resources/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Diagrams and images
+
+- [ros2_control diagrams](ros2_control_ur_driver.drawio) - `.drawio` file, Author: [Denis Stogl](mailto:denis.stogl@picknik.ai) (PickNik Robotics)
+- [generated_images](generated_images/) - see individual files
+
+
+## 2021-10 ROS World presentation
+
+[Presentation: Making a robot ROS 2 powered - a case study using the UR manipulators](2021-10_ROS_World_2021_Making_a_robot_ROS2_powered.pdf)
```
nbbrooks, line 14: Looks like the pdf compressed down fairly nicely to <5mb. @urrsk would you like this added to the repo?

@destogl this could get updated in the future, maybe we should set up git lfs for .pdfs

ur_robot_driver/doc/resources/README.md
```diff
@@ -0,0 +1,30 @@
+# Resources about ROS 2 Universal Robot driver
+
+The files in this folder are available for use under [![License](https://i.creativecommons.org/l/by/4.0/88x31.png)](http://creativecommons.org/licenses/by/4.0/) license.
+
+
+## Diagrams and images
+
+- [ros2_control diagrams](ros2_control_ur_driver.drawio) - `.drawio` file, Author: [Denis Stogl](mailto:denis.stogl@picknik.ai) (PickNik Robotics)
```
nbbrooks, line 8: I tried download the .drawio file and loading it in https://app.diagrams.net/ and got this error

```
Error loading file
Not a diagram file (error on line 15 at column 76: Specification mandates value for attribute crossorigin)
```

Am I doing something wrong? I am not familiar with drawio files.

ur_robot_driver/doc/resources/README.md
urrsk: Why Creative Common license?

ur_robot_driver/doc/resources/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+*Presenter: Denis Štogl*
+
+Authors:
+
+  - Denis Štogl (PickNik Inc.)
+  - Dr. Nathan Brooks (PickNik Inc.)
+  - Lovro Ivanov (PickNik Inc.)
+  - Dr. Andy Zelenak (PickNik Inc.)
+  - Rune Søe-Knudsen (Universal Robots)
```
urrsk, line 30:
```suggestion
  - Rune Søe-Knudsen (Universal Robots A/S)
```

ur_robot_driver/doc/resources/README.md
urrsk: The size is ok

urrsk (changes requested):
@destogl Thanks for this pull request
I would love if we could make a gif-animation of videos in the presentation, at least the one showing the speed scaling.
I do not think we should have a specific license for this folder.

ur_robot_driver/doc/resources/README.md
destogl: The idea is to encourage people to use the information from the presentation, but they have to name us on each use.

ur_robot_driver/doc/resources/README.md
destogl: I just tested it and it works fine.

How did you download the file? I used `wget` and a link to the raw file. Than drag-and-droped it to the diagrams.net

ur_robot_driver/doc/resources/README.md
destogl, line 3:
```suggestion
The files in this folder are available for use under BSD-3-Clause license.
```

urrsk (approved):

destogl:
@nbbrooks @urrsk please check one more time the result of this PR, especially:

- [main README](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/add-materials-from-ros-world/README.md)
- [resources README](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/add-materials-from-ros-world/ur_robot_driver/doc/resources/README.md) - referenced in the main README

Two questions:

1. I don't like very much that GitHub cannot show videos, but they have to be downloaded. Should I generate high-quality gifs instead?
2. Should be set up [`git lfs`](https://git-lfs.github.com/) so that not everybody who is cloning the repository clones also all this big files? It is simply unnecessary.

urrsk:
> 

@destogl Good question. Maybe we should place them on e.g. youtube instead?

ur_robot_driver/doc/resources/README.md
```diff
@@ -0,0 +1,37 @@
+# Resources about ROS 2 Universal Robot driver
```
nbbrooks, line 1:
```suggestion
# Resources about the ROS 2 Universal Robot driver
```

nit

nbbrooks (approved):
Tiny nit, otherwise let's merge once Sean update's youtube video descriptions. Thanks!

edit: And remove video files

fmauch (changes requested):
I think the video files and gif files should still be removed from this PR, right_

nbbrooks:
> I think the video files and gif files should still be removed from this PR, right_

Good catch!

destogl:
> I think the video files and gif files should still be removed from this PR, right_

@fmauch: I would keep the added gif since it is added to the mail README.md to make it more attractive. IMO the gif represents very good what driver is capable of: Scaled control and MoveIt. Do you think we can keep it there or should it be removed?

fmauch:
> > I think the video files and gif files should still be removed from this PR, right_
> 
> @fmauch: I would keep the added gif since it is added to the mail README.md to make it more attractive. IMO the gif represents very good what driver is capable of: Scaled control and MoveIt. Do you think we can keep it there or should it be removed?

I'm not a huge fan of adding ~25MB of binary data (gif, PDFs, pngs) to the repo. I agree that the gif is a nice-to-have and I don't want to block this. I just would like to point out my bad stomach feeling with this.

Just to consider: Should we maybe move that to another repo such as https://github.com/UniversalRobots/Universal_Robots_ROS_Tutorials? I wouldn't be so space-conservative on a pure documentation repository.

On the other hand, this won't affect too many people as soon as we would have binary releases in place...

urrsk:
The motivation for uploading the videos to youtube was to remove them from the repo. I will suggest what we remove them.

destogl:
> The motivation for uploading the videos to youtube was to remove them from the repo. I will suggest what we remove them.

The videos are removed. I only left gif showing driver's capabilities in the main README. It makes the whole repo more vivid. I can remove those too, as well as the pdf with presentation and also documentation figures. But then the whole PR doesn't make sense anymore.

@urrsk and @fmauch please decide if we should move this into another repository or if it should stay here.

fmauch:
@destogl what is the benefit of keeping the figures  separately? As far as I can se, they are included in the PDFs already, right?

fmauch:
I don't really see the benefits of adding the generated images as they are already included into the PDFs and can be extracted.

fmauch:
@destogl I made some minor changes and I think it is ready to me merged like that.

@nbbrooks I took the video descriptions from your youtube videos, that's why I re-asked your approval of this.

fmauch (approved):
I think this is a rather good level of binary data.

destogl:
> I think this is a rather good level of binary data.

Thanks for keeping up with the review!

PR #227, opened 2021-10-26
Empirically set update rates.

livanov93:


gavanderhoorn:
Tbh I'm slightly confused (and surprised) there are so many fundamental differences between the ROS 1 and ROS 2 versions of this driver.

I understand there are changes on the ROS side of things, but the layer between the `urcl` and that ROS side seems like it would be a good candidate for reuse. Perhaps my impression is wrong, but it seems that reuse opportunity is not exploited fully.

livanov93:
> > Can you share some experiece from the ROS1 version of the driver. I can see the [update loop](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/ec2beb65afd6be5373d1e21a86dd42a07e25b447/ur_robot_driver/src/hardware_interface_node.cpp#L139-L160) is run as fast as possible.
> 
> "as fast as possible" is not how I would describe it.
> 
> Yes, that's what the comment says, but it really means that the UR controller dictates the control flow of the hardware interface.
> 
> In other words: data coming in from the UR controller is consumed as fast as possible (or: reads are blocking (with a timeout) and will only return when new data arrives). This immediately means that the `hardware_interface` "runs" at the internal control rate of the UR controller.
> 
> To me that seems like the ideal situation, as it keeps everything in-sync.

Regarding the colloquial expression "as fast as possible" I am aware of what it means and there is already a PR: ros-controls/ros2_control#568 that addresses this situation where the blocking calls to the hardware actually dictate the rate.



> Tbh I'm slightly confused (and surprised) there are so many fundamental differences between the ROS 1 and ROS 2 versions of this driver.
> 
> I understand there are changes on the ROS side of things, but the layer between the `urcl` and that ROS side seems like it would be a good candidate for reuse. Perhaps my impression is wrong, but it seems that reuse opportunity is not exploited fully.

Differences are coming from the changes in ros2_control in comparison to ros1_control. The low level components (like the driver ) are currently implementations of `rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface` and are actually out of the usual node's ability to publish/subscribe/serve clients and the intent is to move all this onto the controllers.

Still, the intention was not to recreate the wheel, and wherever there was possibility of reuse, we seized it. That can be seen from the author marks and licenses within the code itself.

PR #229, opened 2021-10-28
Don't use sleep in update loop

livanov93:
Depends on https://github.com/ros-controls/ros2_control/pull/568


destogl:
now depends on #263

livanov93:
> now depends on #263

It is more  #263 replaced this one!

PR #231, opened 2021-10-28
Add ft_sensor frame.

destogl:
This adds a separate frame for tf sensor values in urdf.

Note: plase wait with merge this is not exactly correct.﻿


fmauch:
I would close this by now. This is rather outdated and was never followed up.

Issue #210, opened 2021-10-18
Pipeline producer overflowed!


kschwan:
Hi,

I'm trying to run the ROS2 driver with a simulated (running in a VM) UR5e 5.11.4.108353. Things seem to run, except the terminal gets flooded with `[UR_Client_Library]: Pipeline producer overflowed! <RTDE Data Pipeline>` error messages.

My host system is Ubuntu 20.04, ROS2 Rolling, and I followed the instructions in the [README](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/README.md).

Here is the terminal output that I'm getting: [terminaloutput.txt](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/files/7368713/terminaloutput.txt)

destogl:
Can you give us joint state topic frequency? Try executing
```
ros2 topic hz /joint_states
```

Thanks!

kschwan:
> @kschwan It is normal that the communication disappears if you use `headless_mode` (requires remote control mode) and suddenly change to local control mode on the teach pendant.
> 
> @GewoonMidas We are working on stabilizing the frequency on the following, with following PRS:
> 
> * [Parametrize sleeping in the update loop ros-controls/ros2_control#568](https://github.com/ros-controls/ros2_control/pull/568)
> * [Don't use sleep in update loop #229](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/229)

OK, thanks :+1:

livanov93:
@kschwan @GewoonMidas Can you confirm the latest changes at UniversalRobots/Universal_Robots_ROS2_Driver@349aa4f711dc416304d1a93a0f89192e9bf1ff8a are working on your part?

GewoonMidas:
@livanov93 I am not sure if I will have time to test it on the UR5e. I will try to run it next week. For this problem #227 solved the issue for me.

kschwan:
> @kschwan @GewoonMidas Can you confirm the latest changes at [349aa4f](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/commit/349aa4f711dc416304d1a93a0f89192e9bf1ff8a) are working on your machines?

I've gotten a new PC, and I've tried running it again. I am now getting the following unrelated error when launching the driver stack. I don't have time to look into it further right now, I'm sorry :/

```
[...]

PR #233, opened 2021-11-02
Make reading and writing work when hardware is disconnected

livanov93:
In the current implementation of the ros-controls/ros2_control [System](https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/src/system.cpp#L222) the [read](https://github.com/ros-controls/ros2_control/blob/c4cdbee6c9139f9e85c9d03d62d05ca683cc83a8/hardware_interface/src/system.cpp#L215-L229) and [write](https://github.com/ros-controls/ros2_control/blob/c4cdbee6c9139f9e85c9d03d62d05ca683cc83a8/hardware_interface/src/system.cpp#L231-L245) are not called if their lifecycle state is different than `PRIMARY_STATE_INACTIVE` or `PRIMARY_STATE_ACTIVE`. This is changed when the `SystemInterface` returns [`hardware_interface::return_type::ERROR`](https://github.com/ros-controls/ros2_control/blob/c4cdbee6c9139f9e85c9d03d62d05ca683cc83a8/hardware_interface/src/system.cpp#L241). The state management when the error is returned can be see [here](https://github.com/ros-controls/ros2_control/blob/c4cdbee6c9139f9e85c9d03d62d05ca683cc83a8/hardware_interface/src/system.cpp#L166-L185). 

This also fixes the `Pipeline producer overflowed!` problem - #210   in non-headless mode when the program is not yet running, but the driver is started.

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -511,6 +511,7 @@ hardware_interface::return_type URPositionHardwareInterface::read()
   }
 
   RCLCPP_ERROR(rclcpp::get_logger("URPositionHardwareInterface"), "Unable to read from hardware...");
+  // TODO(anyone): could not read from the driver --> return ERROR --> on error will be called
   return hardware_interface::return_type::OK;
 }
 
@@ -535,6 +536,7 @@ hardware_interface::return_type URPositionHardwareInterface::write()
   }
 
   RCLCPP_ERROR(rclcpp::get_logger("URPositionHardwareInterface"), "Unable to write to hardware...");
+  // TODO(anyone): could not read from the driver --> return ERROR --> on error will be called
   return hardware_interface::return_type::OK;
 }
 
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -510,8 +510,8 @@ hardware_interface::return_type URPositionHardwareInterface::read()
     return hardware_interface::return_type::OK;
   }
 
-  // TODO(anyone): could not read from the driver --> reset controllers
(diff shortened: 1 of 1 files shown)


ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -510,8 +510,8 @@ hardware_interface::return_type URPositionHardwareInterface::read()
     return hardware_interface::return_type::OK;
   }
 
-  // TODO(anyone): could not read from the driver --> reset controllers
-  return hardware_interface::return_type::ERROR;
+  RCLCPP_ERROR(rclcpp::get_logger("URPositionHardwareInterface"), "Unable to read from hardware...");
+  return hardware_interface::return_type::OK;
```
destogl, line 514:
```suggestion
  // TODO(anyone): could not read from the driver --> return ERROR --> on error will be called
  return hardware_interface::return_type::OK;
```

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -534,7 +534,8 @@ hardware_interface::return_type URPositionHardwareInterface::write()
     return hardware_interface::return_type::OK;
   }
 
-  return hardware_interface::return_type::ERROR;
+  RCLCPP_ERROR(rclcpp::get_logger("URPositionHardwareInterface"), "Unable to write to hardware...");
+  return hardware_interface::return_type::OK;
```
destogl, line 538:
```suggestion
  // TODO(anyone): could not read from the driver --> return ERROR --> on error will be called
  return hardware_interface::return_type::OK;
```

destogl (approved):
Looking good! What do you think to leave "TODOs" for the future development to do proper error handling.

PR #237, opened 2021-11-05
Transform force-torque into tcp frame.

### LINKED ISSUE #235: Force/torque sensor broadcaster publishes wrench with incorrect orientation
Expected: Wrench measurement from e-Series wrist force/torque sensor is relative to whatever frame ID I set when configuring the force/torque controller plugin (for example, the `tool0` frame at the robot tool flange).

Actual: Force/torque wrench measurements are oriented relative to the robot's base link, and are consequently almost always incorrect as the robot's pose changes.

I think this is just because the UR driver's convention for reporting TCP force is to report it relative to the base coordinate system, so additional work is needed to transform it into the correct frame. This is reflected in [this guide](https://www.universal-robots.com/articles/ur/programming/weight-detection-of-any-picked-object-using-e-series/) about using a UR script program to weigh an object using the force/torque sensor.

Additional potentially-useful info:
- ROS2 Rolling
- UR ROS2 driver package commit: `349aa4f`
- UR client library package commit: `90338a7` (a few commits later than the `1.0.0` tag).
- UR Software version `5.11.0.108249`

livanov93:
Closes #235 
Sets the rotation of the wrench message into the tool center point.


destogl (approved):
This looks good. Please test it on hardware and then it can be merged.

gavanderhoorn:
Isn't this basically a copy from the ROS 1 driver?

destogl:
> Isn't this basically a copy from the ROS 1 driver?

Yes. Shouldn't it be?

schornakj:
@destogl I tested this on my robot and the wrench measurements are now correctly published relative to the `tool0` frame that was specified when configuring the force/torque controller.

I'm testing this using:
```
ros2 launch ur_bringup ur_control.launch.py headless_mode:=True use_fake_hardware:=False fake_sensor_commands:=False robot_ip:="[ip_address]" ur_type:="ur5e" launch_rviz:=True
```

gavanderhoorn:
> > Isn't this basically a copy from the ROS 1 driver?
> 
> Yes. Shouldn't it be?

there was no mention of that anywhere. Not in the commit, nor in the PR description.

Seeing as it was already part of the ROS 1 driver, and the ROS 2 version here is a continuation of that development, but this functionality somehow got removed, it would be a good idea to add a note where this came from to the commit. As-is, it's almost like this functionality is new, while in reality is was (inadvertently likely) removed.

Among other things, noting where it came from and that it's reinstated functionality would link this back to the ROS 1 driver, connecting it to the history there as well, which may help in the future with maintenance tasks.

destogl:
@livanov93 can you please add link to the ROS1 driver when squash-merging into commit?

Also, it would make sense to add comment in the code too. Then you can merge this since @schornakj confirmed the functionality.

livanov93:
@destogl @gavanderhoorn I have added the the original source of the code in the comments.  When merging I will leave another link to ROS1 driver.

PR #239, opened 2021-11-10
Move Servo launching into the main MoveIt launch file. Make it optional.

AndyZe:
The realtime control launch file was previously off in the `ur_moveit_config` folder, all by itself like an orphan. I think we should include it with the launching of other MoveIt stuff.

This also reduces duplicate code in the two launch files.

This also adds a loaded-but-not-started forward_position_controller which is more suitable for realtime control than a JointTrajectoryController.

Test like this:

`ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=false`

`ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:="xxx.xxx" use_fake_hardware:=true launch_rviz:=true`

`ros2 control switch_controllers --start forward_position_controller --stop joint_trajectory_controller`

`ros2 service call /servo_node/start_servo std_srvs/srv/Trigger {}`

Then I use this package to send simple test commands, to move in a circle:  https://github.com/AndyZe/simple_servo_twist_pub


AndyZe:
This should be useful for people like @MarqRazz and myself who are testing with UR robots.

ur_bringup/launch/ur_moveit.launch.py
```diff
@@ -355,5 +369,8 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument("launch_rviz", default_value="true", description="Launch RViz?")
     )
+    declared_arguments.append(
+        DeclareLaunchArgument("launch_servo", default_value="true", description="Launch Servo?")
```
AndyZe, line 373: I made it so Servo is launched by default. If the user doesn't need it, it sits quietly, waiting for a service call to activate.

AndyZe, line 373: I made it so Servo is launched by default. If the user doesn't need it, it sits quietly, waiting for a service call to activate.

We could change the default to false, if people feel strongly.

ur_moveit_config/config/ur_servo.yaml
```diff
@@ -19,13 +19,16 @@ publish_period: 0.004  # 1/Nominal publish rate [seconds]
...
+command_out_type: trajectory_msgs/JointTrajectory
 
 # What to publish? Can save some bandwidth as most robots only require positions or velocities
-publish_joint_positions: false
+publish_joint_positions: true
 publish_joint_velocities: true
 publish_joint_accelerations: false
 
+## Plugins for smoothing outgoing commands
+smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
```
AndyZe, line 30: This is required by a change within MoveIt.

ur_moveit_config/config/ur_servo.yaml
```diff
@@ -53,7 +56,7 @@ cartesian_command_in_topic: ~/delta_twist_cmds  # Topic for incoming Cartesian t
 joint_command_in_topic: ~/delta_joint_cmds # Topic for incoming joint angle commands
 joint_topic: /joint_states
 status_topic: ~/status # Publish status to this topic
-command_out_topic: /forward_velocity_controller/commands # Publish outgoing commands here
+command_out_topic: /joint_trajectory_controller/joint_trajectory # Publish outgoing commands here
```
AndyZe, line 59: This is the default controller type that is launched from this package at this point in time. It works fine with Servo.

AndyZe:
Pinging to get this merged. It reduces launch file duplication quite a bit.

Does the "coverage build" CI job fail regularly or is it something I should fix?

AndyZe:
A very boring video of moving in a sine wave, just to prove this works. This uses the Python script [here](https://github.com/AndyZe/simple_servo_twist_pub/blob/main/simple_servo_twist_pub/publisher_member_function.py) to send a sine wave command.

https://user-images.githubusercontent.com/11284393/144480219-ee775cf0-7847-49bd-9040-e93ed9a34206.mp4

AndyZe:
A video from testing on a real UR5e. Again, using the Python script [here](https://github.com/AndyZe/simple_servo_twist_pub/blob/main/simple_servo_twist_pub/publisher_member_function.py) to send a sine wave command.

https://user-images.githubusercontent.com/11284393/144669481-bb9c2d34-3603-4d39-9ab6-c3ba422b35f8.mp4

ur_bringup/config/ur_controllers.yaml
```diff
@@ -107,3 +110,14 @@ forward_velocity_controller:
...
+forward_position_controller:
+  ros__parameters:
+    joints:
+      - shoulder_pan_joint
+      - shoulder_lift_joint
+      - elbow_joint
+      - wrist_1_joint
+      - wrist_2_joint
+      - wrist_3_joint
+    interface_name: position
```
destogl, line 123: If you are using the specific controller, this parameter is not needed.

```suggestion
```

ur_bringup/launch/ur_moveit.launch.py
```diff
@@ -41,6 +41,7 @@ def launch_setup(context, *args, **kwargs):
     use_fake_hardware = LaunchConfiguration("use_fake_hardware")
     fake_sensor_commands = LaunchConfiguration("fake_sensor_commands")
     launch_rviz = LaunchConfiguration("launch_rviz")
+    launch_rviz = LaunchConfiguration("launch_servo")
```
destogl, line 44: You probably meant something like this:
```suggestion
    launch_servo = LaunchConfiguration("launch_servo")
```

ur_bringup/launch/ur_moveit.launch.py
```diff
@@ -243,12 +244,25 @@ def launch_setup(context, *args, **kwargs):
...
-        mongodb_server_node,
-        rviz_node,
-        static_tf,
-    ]
+    # Servo node for realtime control
+    servo_yaml = load_yaml("ur_moveit_config", "config/ur_servo.yaml")
+    servo_params = {"moveit_servo": servo_yaml}
+    servo_node = Node(
+        package="moveit_servo",
+        condition=IfCondition(launch_rviz),
```
destogl, line 252:
```suggestion
        condition=IfCondition(launch_servo),
```

ur_bringup/launch/ur_moveit.launch.py
destogl: I agree to start servo per default.

destogl (changes requested):
I have a few rather small comments. Although in general I agree with this PR, I have few questions to clarify before merge:

I understand that servo is currently using position controller instead of velocity controller. What is the reason behind this and would it be possible to parameterize this? I can imagine that using velocity FFWD controller the movements could be smoother.

MarqRazz:
> I understand that servo is currently using position controller instead of velocity controller. What is the reason behind this and would it be possible to parameterize this? I can imagine that using velocity FFWD controller the movements could be smoother.

The position control of Servo is buttery smooth (actually smoother than jogging with the builtin UR controller because it jerks near singularities). I would prefer that we get position control working properly before we jump to velocity control (i.e. Servo can accurately calculate joint velocities based on its position command so that we can perform clamping and that the tool frame moves at the speed commanded by the user). All arms are not guaranteed to have an interface to control velocity so I suggest we make position control a first class supported feature.

AndyZe:
I actually switched to position control because velocity control isn't supported with the ros2 generic hw simulation yet. But Marq makes some good points too

ur_bringup/launch/ur_moveit.launch.py
AndyZe: thanks!

destogl (approved):
This looks good to me! Thanks!

PR #243, opened 2021-11-16
Fix 🛠: correct launch file arguments in files for individual robots.

destogl:


livanov93 (approved):
LGTM! We can see the community response and then potentially rename it again.

PR #241, opened 2021-11-15
Remove ignition as required dependency

livanov93:
Partially addressing #240 after #232 . @vatanaksoytezer please import ignition as it is discussed in #240 in follow-up PR.
Removes the ignition merged PR, as it is concluded it makes Ignition as required dependency for bare-bone driver.

Issue #240, opened 2021-11-11
Ignition from required to optional dependency


livanov93:
We have to enable people to use bare-bone driver.
We should make Ignition as optional dependency for user that aim to do work with that simulator. Currently it is required dependency, added into main `ur_control.launch.py` file, which can't be run if ignition is not installed and the ros ignition wrapper is built in the workspace.
TODO:

- add .repos file for users that want to use ignition (see example for moveit within the driver repo)
- add additional documentation for build/install procedure for ignition users, current links are dead
- add separate launch file, don't mix it into ur_control.launch.py (include it, but don't change it)

@vatanaksoytezer please address theses changes after the work you have put into #232 
Thanks!

fmauch:
I think, this can be close by now.

PR #245, opened 2021-11-23
Add support for Gazebo and Ignition simulation

destogl:
**NOTE:** This PR will be soon moved to another repository to separate simulation dependencies.

Changes needed to run Gazebo and Ignition simulation backed-up by ros2_control. 
Currently, the plugins are not released, so you have to compile them from source.
Check simulations.repos file for this.

Some left TODOs before merging:
- [ ] Add documentation about this into README.md
- [ ] wati for "initial position"-functionality get merged into simulators:
  - [ ] ros-simulation/gazebo_ros2_control#100
  - [ ] Igntion - TBD
- [ ] Resolve URDF changes to enable visuals in Gazebo (and test this with Ignition)

## Usage:
Tested robot (default) is ur5e - although it should also work for other robots.

Checkout repos from `simulations.repos` file.

After launching simulation, you can test the robot's behavior using test script for joint trajectory controller.
Package ros2_control_demos is needed for this.

```
ros2 launch ur_bringup test_joint_trajectory_controller.launch.py
```

#### Ignition:

```
ros2 launch ur_bringup ur_sim_control.launch.py
```

### Gazebo:
Add `gazebo_ros2_control` repository on PR#100 (ros-controls/gazebo_ros2_control#100).

```
ros2 launch ur_bringup ur_sim_control.launch.py gazebo_sim:=true ignition_sim:=false
```

Note: this PR containes changes from #231.


livanov93:
@destogl I would like to keep parametrization from the YAML file because it is one of the many things that were done in ros1, and are quite nice to have. So gazebo has problems with parsing those paths? Is it possible to solve this in any other way, while keeping the YAML paths propagation?

livanov93:
@destogl I would also like to create a separate gazebo_control.ros2_control.xacro file to avoid having one with many options.

ur_description/urdf/ur_macro.xacro
```diff
@@ -109,13 +116,13 @@
       <visual>
         <origin xyz="0 0 0" rpy="0 0 ${pi}"/>
         <geometry>
-          <mesh filename="${base_visual_mesh}"/>
+          <mesh filename="file://$(find ur_description)/meshes/${base_visual_mesh}"/>
```
destogl, line 119: Maybe something like this?

```suggestion
          <mesh filename="file://$(find ur_description)/meshes/${ur_type}/visual/base.dae"/>
```

ur_description/urdf/ur.ros2_control.xacro
```diff
@@ -257,6 +261,70 @@
...
+    <!-- Gazebo plugins -->
+    <gazebo reference="world">
+    </gazebo>
+    <gazebo>
+      <plugin filename="libgazebo_ros2_control.so" name="gazebo_ros2_control">
+        <parameters>${gazebo_controllers}</parameters>
+      </plugin>
+    </gazebo>
+
+    <!-- TODO: Implement initial positions support in Gazebo - check ros-controls/gazebo_ros2_control#100 -->
```
relffok, line 276: This colon breaks my launch using the testing repos of galactic. I didn't dig deeper but removing it made launching possible. One older issue I found: https://github.com/ros2/launch_ros/issues/136

ur_bringup/launch/ur_gazebo_sim_control.launch.py
```diff
@@ -0,0 +1,366 @@
...
+
+    initial_joint_controllers = PathJoinSubstitution(
+        [FindPackageShare(runtime_config_package), "config", controllers_file]
+    )
+
+    rviz_config_file = PathJoinSubstitution(
+        [FindPackageShare(description_package), "rviz", "view_robot.rviz"]
+    )
+
+    robot_description_content = Command(
```
MarqRazz, line 66: @destogl do you have an example launch command I could test this out with? Do we really need to manually specify this many parameters to bring the robot up in Ignition?

MarqRazz:
@destogl after getting this working I was surprised that this is using Gazebo classic!?? 

I need this working in Ignition, so if you know of any conflicts this might have as I get the newer version going again please feel free to tag me.

MarqRazz:
Wont this PR have the same issues (#240) that @vatanaksoytezer's PR had? This will make the repo depend on Gazebo classic and if I also depend on Ignition then users will need to install both. @destogl do you have a plan on getting around this here?

relffok:
> Wont this PR have the same issues (#240) that @vatanaksoytezer's PR had? This will make the repo depend on Gazebo classic and if I also depend on Ignition then users will need to install both.

I don't see a specific dependency to gazebo here. The `<gazebo>` tags in the urdf won't get parsed if not by gazebo. The build and the other launch files like `ur_control.launch.py` should work. Only `ur_gazebo_sim_control.launch.py` will fail because of gzclient/gzserver not being available. 

Though it looks like igniton is still a dependency, since the main branch was not merged yet including the https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/241 revert commit.

vatanaksoytezer:
> Though it looks like igniton is still a dependency, since the main branch was not merged yet including the #241 revert commit.

Ignition PR literally lived around a week and reverted, I don't understand how can it be dependency, while it's never pushed rosdistro and immediately reverted ...

relffok:
> Ignition PR literally lived around a week and reverted, I don't understand how can it be dependency, while it's never pushed to rosdistro and immediately reverted on main branch ...

Seems like gazebo-simulation branch was forked from main right that time, therefore the revert was not included yet. 

Anyway, I removed the merge conflict + colon bug in a separate PR https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/249 to your branch, if you can have a look @destogl .

MarqRazz:
> I don't see a specific dependency to gazebo here. The `<gazebo>` tags in the urdf won't get parsed if not by gazebo. The build and the other launch files like `ur_control.launch.py` should work. Only `ur_gazebo_sim_control.launch.py` will fail because of gzclient/gzserver not being available.

Its more a question of if you want the package.xml to include the actual dependencies or if its the users responsibility to go and find them. The ROS1 robot_description [depends on Gazebo classic packages](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_gazebo/package.xml#L25-L26).

gavanderhoorn:
> Its more a question of if you want the package.xml to include the actual dependencies or if its the users responsibility to go and find them. The ROS1 robot_description [depends on Gazebo classic packages](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_gazebo/package.xml#L25-L26).

No, it actually doesn't.

You're linking to the Gazebo package, which was specifically introduced to separate the Gazebo dependencies from the regular `xacro:macro`.

Both the Kinetic nor the [Melodic](https://github.com/ros-industrial/universal_robot/blob/9d02ccadc8fb9b5cfeaf0d5b0c0c7708d2461447/ur_description/package.xml#L21-L29) version of `ur_description` do not depend on any simulator.

MarqRazz:
> No, it actually doesn't.
> 
> You're linking to the Gazebo package, which was specifically introduced to separate the Gazebo dependencies from the regular `xacro:macro`.
> 
> Neither the Kinetic nor the [Melodic](https://github.com/ros-industrial/universal_robot/blob/9d02ccadc8fb9b5cfeaf0d5b0c0c7708d2461447/ur_description/package.xml#L21-L29) version of `ur_description`depends on any simulator.

My bad, I have been doing everything from source lately and look at the entire repo. So is it correct to say that it requires developers have everything installed if we supported Gazebo Classic and Ignition?

This still doesn't seem like it addresses the fact that this PR depends on `gazebo_ros2_control`, `Gazebo Classic` and that the launch file is currently located in `ur_bringup`. Where would you say those dependencies belong? Should this be split into its own package like the ROS1 `ur_gazebo`?

gavanderhoorn:
> I have been doing everything from source lately and look at the entire repo

it's true at the repository level those dependencies of course exist, but without hosting every package in its own repository, that's always going to be the case (and one-package-per-repository does not really scale).

The suggested workflow when building from source would be to specifically *ignore* packages you're not interested in. So use either the white/blacklisting feature of `catkin`, use `(CATKIN|COLCON|AMENT)_IGNORE` files or (really a last resort) remove the packages you don't want.

It's not ideal, but at least this way it's feasible to build subsets of packages from the same repository and not always drag in heavy dependencies.

(incidently, `industrial_ci` supports this in a nice way: configure it to clone an "upstream repository" providing some dependency, and then tell it to prune certain packages before installing dependencies of your dependency)

> This still doesn't seem like it addresses the fact that this PR depends on `gazebo_ros2_control`, `Gazebo Classic` 

I agree. I just wanted to correct your statement about `ros-industrial/universal_robot/ur_description` depending on Gazebo.

> and that the launch file is currently located in `ur_bringup`. Where would you say those dependencies belong? Should this be split into its own package like the ROS1 `ur_gazebo`?

I would suggest to not use `ur_bringup` for *everything*.

That, combined with separating packages (ie: responsibilities) as much as possible should allow for relatively finegrained management of dependencies.

livanov93:
@MarqRazz @vatanaksoytezer This PR is still a work in progress and it will be reviewed a few times before merging.
IMO The approach to be able to separate simulation (gazebo/ignition) users from the users of bare-bone driver consist of a few points:

- if other dependencies are needed create .repos file that can import necessary dependencies from the source
- if .repos file is not an option consider using `git submodule option` that will link an empty package/folder if the package  (in this situation `ur_gazebo`) is not needed
- move everything gazebo/ignition-related (except description stuff into) `ur_gazebo` which should be an external package (as mentioned above .repos or git submodule approach) - move also launch files which can be an extended set of main `ur_control.launch.py` file for barebone driver - we shouldn't add too many flags/options into one launch file because it creates confusion and it will eventually make the whole launch file system hard to maintain
- add separate ur_macro.xacro (e.g. ur_gazebo(ignition)_macro.xacro) in `ur_description` which will support all the necessary gazebo/ignition references and will solve the problem of parsing mesh file paths (maybe this additional urdf/xacro file  can also be part of ur_gazebo )
- of course, README file to document all the necessary steps to use the extended option of the driver package - in this case gazebo/ignition

destogl:
There is new a fixed proposal also including Ignition simulation - because those two are very, very similar.

ur_bringup/config/ur_controllers.yaml
```diff
@@ -1,5 +1,6 @@
 controller_manager:
   ros__parameters:
+    update_rate: 500  # Hz
```
destogl, line 3: Remove this before merge.

ur_bringup/config/test_goal_publishers_config.yaml
```diff
@@ -47,7 +47,7 @@ publisher_joint_trajectory_controller:
       - wrist_2_joint
       - wrist_3_joint
 
-    check_starting_point: true
+    check_starting_point: false
```
destogl, line 50: Revert this before merge.

MarqRazz:
I have tested the PR and can confirm the Gazebo classic simulation works but when I run Ignition I get the following errors and the UI crashes. I will keep digging and let you know if I can find the root error.

I have other Ignition worlds that I can run and get the same `libEGL warning: DRI2: failed to create dri screen` but everything launches and works as expected. Not sure why I am getting the duplicate proto message, but that is not output in my other test cases???
```
[ign gazebo-5] [libprotobuf ERROR google/protobuf/descriptor_database.cc:58] File already exists in database: ignition/msgs/actor.proto
[ign gazebo-5] [libprotobuf FATAL google/protobuf/descriptor.cc:1358] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size): 
[ign gazebo-5] terminate called after throwing an instance of 'google::protobuf::FatalException'
[ign gazebo-5]   what():  CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size): 
[ign gazebo-5] Stack trace (most recent call last):
[ign gazebo-5] #31   Object "/usr/lib/x86_64-linux-gnu/ign-gazebo-6/plugins/libignition-gazebo-user-commands-system.so", at 0x7f220855bdb2, in ignition::gazebo::v6::systems::CreateCommand::Execute()
[ign gazebo-5] #30   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224afc1c1, in ignition::gazebo::v6::SdfEntityCreator::CreateEntities(sdf::v12::Model const*)
[ign gazebo-5] #29   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224b10130, in void ignition::gazebo::v6::EventManager::Emit<ignition::common::EventT<void (unsigned long, std::shared_ptr<sdf::v12::Element>), ignition::gazebo::v6::events::LoadPluginsTag>, unsigned long const&, std::shared_ptr<sdf::v12::Element> const&>(unsigned long const&, std::shared_ptr<sdf::v12::Element> const&)
[ign gazebo-5] #28   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224b2aa09, in ignition::gazebo::v6::SimulationRunner::LoadPlugins(unsigned long, std::shared_ptr<sdf::v12::Element> const&)
[ign gazebo-5] #27   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224b2994b, in ignition::gazebo::v6::SimulationRunner::LoadPlugin(unsigned long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<sdf::v12::Element> const&)
[ign gazebo-5] #26   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224b45e58, in ignition::gazebo::v6::SystemLoader::LoadPlugin(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<sdf::v12::Element> const&)
[ign gazebo-5] #25   Object "/lib/x86_64-linux-gnu/libignition-gazebo6.so.6", at 0x7f2224b4676c, in 
[ign gazebo-5] #24   Object "/lib/x86_64-linux-gnu/libignition-plugin1-loader.so.1", at 0x7f2222134670, in ignition::plugin::Loader::LoadLib(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[ign gazebo-5] #23   Object "/lib/x86_64-linux-gnu/libignition-plugin1-loader.so.1", at 0x7f2222133a7d, in ignition::plugin::Loader::Implementation::LoadLib(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[ign gazebo-5] #22   Object "/lib/x86_64-linux-gnu/libdl.so.2", at 0x7f22290c23d9, in dlopen
[ign gazebo-5] #21   Object "/lib/x86_64-linux-gnu/libdl.so.2", at 0x7f22290c2b58, in 
[ign gazebo-5] #20   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22292de982, in _dl_catch_error
[ign gazebo-5] #19   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22292de8b7, in _dl_catch_exception
[ign gazebo-5] #18   Object "/lib/x86_64-linux-gnu/libdl.so.2", at 0x7f22290c234b, in 
[ign gazebo-5] #17   Object "/lib64/ld-linux-x86-64.so.2", at 0x7f22297025f9, in 
[ign gazebo-5] #16   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22292de8b7, in _dl_catch_exception
[ign gazebo-5] #15   Object "/lib64/ld-linux-x86-64.so.2", at 0x7f22297030be, in 
[ign gazebo-5] #14   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22292de914, in _dl_catch_exception
[ign gazebo-5] #13   Object "/lib64/ld-linux-x86-64.so.2", at 0x7f22296fec90, in 
[ign gazebo-5] #12   Object "/lib64/ld-linux-x86-64.so.2", at 0x7f22296feb89, in 
[ign gazebo-5] #11   Object "/lib/x86_64-linux-gnu/libignition-msgs5.so.5", at 0x7f22013d3768, in void std::call_once<void (&)()>(std::once_flag&, void (&)())
[ign gazebo-5] #10   Object "/lib/x86_64-linux-gnu/libpthread.so.0", at 0x7f222916847e, in 
[ign gazebo-5] #9    Object "/lib/x86_64-linux-gnu/libignition-msgs5.so.5", at 0x7f22013ce3ed, in protobuf_ignition_2fmsgs_2factor_2eproto::AddDescriptorsImpl()
[ign gazebo-5] #8    Object "/lib/x86_64-linux-gnu/libprotobuf.so.17", at 0x7f2223e8387d, in 
[ign gazebo-5] #7    Object "/lib/x86_64-linux-gnu/libgcc_s.so.1", at 0x7f22257485a9, in _Unwind_Resume
[ign gazebo-5] #6    Object "/lib/x86_64-linux-gnu/libgcc_s.so.1", at 0x7f2225747bee, in 
[ign gazebo-5] #5    Object "/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f22257fbd20, in __gxx_personality_v0
[ign gazebo-5] #4    Object "/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f22257fb368, in 
[ign gazebo-5] #3    Object "/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f22257fc38b, in 
[ign gazebo-5] #2    Object "/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f22257f0910, in 
[ign gazebo-5] #1    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22291a0858, in abort
[ign gazebo-5] #0    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f22291c118b, in gsignal
[ign gazebo-5] Aborted (Signal sent by tkill() 68789 0)
[ign gazebo-5] [GUI] [Wrn] [Application.cc:669] [QT] file::/Gazebo/GazeboDrawer.qml:147:3: QML Dialog: Binding loop detected for property "implicitHeight"
[ign gazebo-5] libEGL warning: DRI2: failed to create dri screen
[ign gazebo-5] libEGL warning: DRI2: failed to create dri screen
```

MarqRazz:
So I found the issue and it was because I don't  have the environment variable `$IGNITION_VERSION` set which are used [here](https://github.com/ignitionrobotics/ign_ros2_control/blob/main/ignition_ros2_control/CMakeLists.txt#L28-L47) and [here](https://github.com/ignitionrobotics/ign_ros2_control/blob/main/ignition_ros2_control/package.xml#L13-L16). 
Is this variable normally setup by Ignition when you install it? I don't see anything about adding it in the repo [README](https://github.com/ignitionrobotics/ign_ros2_control)
After setting `export IGNITION_VERSION=fortress` the simulation works!!
ur_ignition

Anyone know why the mesh colors are not being used? In @vatanaksoytezer port they were working.

ur_description/config/ur5e/visual_parameters.yaml
```diff
@@ -1,67 +1,67 @@
 mesh_files:
   base:
     visual:
-      mesh: "package://ur_description/meshes/ur5e/visual/base.dae"
+      mesh: "ur5e/visual/base.dae"
       material:
         name: "LightGrey"
         color: "0.7 0.7 0.7 1.0"
```
MarqRazz, line 7: Why are we specifying the link material here? The .dae files are already colored.

I I delete the `material` variables from this file and `ur_common.xacro` and `ur_macro.xacro` then the robot appears in simulation and Rviz as expected.
image

ur_description/config/ur5e/visual_parameters.yaml
```diff
@@ -1,67 +1,67 @@
...
     visual:
-      mesh: "package://ur_description/meshes/ur5e/visual/upperarm.dae"
+      mesh: "ur5e/visual/upperarm.dae"
       material:
         name: "LightGrey"
         color: "0.7 0.7 0.7 1.0"
     collision:
-      mesh: "package://ur_description/meshes/ur5e/collision/upperarm.stl"
+      mesh: "ur5e/collision/upperarm.stl"
       mesh_files:
```
MarqRazz, line 28: What is the line for? Its only listed for this link.

ur_description/config/ur5e/visual_parameters.yaml
```diff
@@ -1,67 +1,67 @@
...
     visual:
-      mesh: "package://ur_description/meshes/ur5e/visual/wrist2.dae"
+      mesh: "ur5e/visual/wrist2.dae"
       material:
         name: "LightGrey"
         color: "0.7 0.7 0.7 1.0"
     collision:
-      mesh: "package://ur_description/meshes/ur5e/collision/wrist2.stl"
+      mesh: "ur5e/collision/wrist2.stl"
     visual_offset: -0.0997
```
MarqRazz, line 57: What are the offsets for? Can we add a comment why they are needed?

MarqRazz (commented):
Great work @destogl thanks for getting Ignition working!

My only complaint about this would be that I have to have Gazebo classic and Ignition installed if I want to simulate. If we could break them into separate packages then the user could choose their poison.

MarqRazz:
I think the clocks might be out of sync, I cant get motion from the arm when running MoveIt. I can see that the robot is publishing `joint_states` but get the following error when attempting to plan after launching:
`ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:="xxx.xxx" use_fake_hardware:=true launch_rviz:=true`

```
[move_group-1] [moveit_move_group_default_capabilities.move_action_capability]: Motion plan was computed succesfully.
[rviz2-3] [move_group_interface]: Planning request complete!
[rviz2-3] [move_group_interface]: time taken to generate plan: 0.0186026 seconds
[move_group-1] [moveit_move_group_default_capabilities.execute_trajectory_action_capability]: Received goal request
[move_group-1] [moveit_move_group_default_capabilities.execute_trajectory_action_capability]: Execution request received
[move_group-1] [moveit.plugins.moveit_simple_controller_manager]: Returned 1 controllers in list
[move_group-1] [moveit_ros.trajectory_execution_manager]: Validating trajectory with allowed_start_tolerance 0.01
[rviz2-3] [move_group_interface]: Execute request accepted
[move_group-1] [moveit_ros.current_state_monitor]: Didn't received robot state (joint angles) with recent timestamp within 1.000000 seconds.
[move_group-1] Check clock synchronization if your are running ROS across multiple machines!
[move_group-1] [moveit_ros.trajectory_execution_manager]: Failed to validate trajectory: couldn't receive full current joint state within 1s
[move_group-1] [moveit_move_group_default_capabilities.execute_trajectory_action_capability]: Execution completed: ABORTED
[rviz2-3] [move_group_interface]: Execute request aborted
[rviz2-3] [move_group_interface]: MoveGroupInterface::execute() failed or timeout reached
```

ur_description/config/ur5e/visual_parameters.yaml
destogl: This is a porting “relict” from ROS 1. We should check this at least for the visuals. I would open a follow-up issue for this since it is not really simulation related. @livanov93 @urrsk what do you think? Would should be done here? Should we ignore materials for visuals in the ur_macro driver?

ur_description/config/ur5e/visual_parameters.yaml
destogl: Good question...

ur_description/config/ur5e/visual_parameters.yaml
destogl: This is the offset of the visual mesh to the coordinate system of the link. We could add a comment probably with the offiset axis here is set.

destogl:
> Is this variable normally setup by Ignition when you install it? I don't see anything about adding it in the repo README. If you would like I can make a PR and add a note like this one for ign_ros2_control.

I have no idea about this. I didn't have to set any environment variable manually.



> I think the clocks might be out of sync, I cant get motion from the arm when running MoveIt. I can see that the robot is publishing `joint_states` but get the following error when attempting to plan after launching: `ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:="xxx.xxx" use_fake_hardware:=true launch_rviz:=true`

Is this related to the simulation of FakeHardware?

MarqRazz:
> I have no idea about this. I didn't have to set any environment variable manually.

I would assume then that you are using `citadel` then because that's what it defaults to. The [instructions](https://github.com/ignitionrobotics/ign_ros2_control/tree/foxy#compile-from-source) for building from source have been updated.

> Is this related to the simulation of FakeHardware?

This is for running the simulation. Do we need to change the launch to run sim? I am launching the following in two terminals.
`ros2 launch ur_bringup ur_sim_control.launch.py`
`ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:="xxx.xxx" use_fake_hardware:=true launch_rviz:=true`
Should I be using `use_fake_hardware:=false`?

destogl:
> This is for running the simulation. Do we need to change the launch to run sim? I am launching the following in two terminals. `ros2 launch ur_bringup ur_sim_control.launch.py` `ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:="xxx.xxx" use_fake_hardware:=true launch_rviz:=true` Should I be using `use_fake_hardware:=false`?

This doesn't make any difference because for the moveit those parameters are irrelevant.

fmauch:
@destogl should we close this? There are separate repositories for simulation:

ignition: https://github.com/UniversalRobots/Universal_Robots_ROS2_Ignition_Simulation
gazebo: https://github.com/UniversalRobots/Universal_Robots_ROS2_Gazebo_Simulation

Issue #264, opened 2021-12-13
Error: Unable to write to hardware...


mehtas73:
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [dashboard_client-2]: process started with pid [14348]
[INFO] [ros2_control_node-1]: process started with pid [14346]
[INFO] [robot_state_publisher-3]: process started with pid [14350]
[INFO] [rviz2-4]: process started with pid [14352]
[INFO] [spawner-5]: process started with pid [14354]
[INFO] [spawner-6]: process started with pid [14356]
[INFO] [spawner-7]: process started with pid [14358]
[INFO] [spawner-8]: process started with pid [14360]
[INFO] [spawner-9]: process started with pid [14362]
[ros2_control_node-1] 1639420075.880662 [0] ros2_contr: using network interface enp1s0 (udp/10.34.94.105) selected arbitrarily from: enp1s0, eno1
[dashboard_client-2] 1639420075.880662 [0] dashboard_: using network interface enp1s0 (udp/10.34.94.105) selected arbitrarily from: enp1s0, eno1
[dashboard_client-2] [INFO] [1639420075.903021759] [UR_Client_Library]: Connected: Universal Robots Dashboard Server
[dashboard_client-2] 
[robot_state_publisher-3] 1639420075.912579 [0] robot_stat: using network interface enp1s0 (udp/10.34.94.105) selected arbitrarily from: enp1s0, eno1
[robot_state_publisher-3] Link base had 0 children
[robot_state_publisher-3] Link base_link_inertia had 1 children
[robot_state_publisher-3] Link shoulder_link had 1 children
[robot_state_publisher-3] Link upper_arm_link had 1 children
[robot_state_publisher-3] Link forearm_link had 1 children
[robot_state_publisher-3] Link wrist_1_link had 1 children
[robot_state_publisher-3] Link wrist_2_link had 1 children
[robot_state_publisher-3] Link wrist_3_link had 1 children
[robot_state_publisher-3] Link flange had 1 children
[robot_state_publisher-3] Link tool0 had 0 children
[robot_state_publisher-3] [INFO] [1639420075.934047110] [robot_state_publisher]: got segment base
[robot_state_publisher-3] [INFO] [1639420075.934214957] [robot_state_publisher]: got segment base_link
[robot_state_publisher-3] [INFO] [1639420075.934232199] [robot_state_publisher]: got segment base_link_inertia
[...]

Issue #262, opened 2021-12-10
Having trouble running the driver on Rolling


fmauch:
When trying to run the driver on ROS2 Rolling I experienced some problems. (See launchfile output below)

My workspace looks like this and is up-to-date (as of writing this):
```
========= Universal_Robots_Client_Library =========
## master...origin/master
========= Universal_Robots_ROS2_Driver =========
## main...origin/main
========= ur_msgs =========
## ros2...origin/ros2
```

When using the driver with ROS2 Foxy (with the `foxy` branch of this driver), everything works out great. I am currently dropping this here, without further investigating it. If it works for everybody else, feel free to close this, but since I just installed ROS Rolling and tested the driver, I guess others might also fall into this problem.

```
$ ros2 launch ur_bringup ur5e.launch.py robot_ip:=192.168.56.101 launch_rviz:=false
[INFO] [launch]: All log files can be found below /home/mauch/.ros/log/2021-12-10-11-43-24-296701-ids-felino-155786
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [dashboard_client-2]: process started with pid [155792]
[INFO] [ros2_control_node-1]: process started with pid [155790]
[INFO] [robot_state_publisher-3]: process started with pid [155794]
[INFO] [spawner-4]: process started with pid [155796]
[INFO] [spawner-5]: process started with pid [155798]
[INFO] [spawner-6]: process started with pid [155800]
[INFO] [spawner-7]: process started with pid [155802]
[INFO] [spawner.py-8]: process started with pid [155804]
[INFO] [spawner-9]: process started with pid [155806]
[dashboard_client-2] [INFO] [1639133005.376358381] [UR_Client_Library]: Connected: Universal Robots Dashboard Server
[dashboard_client-2]
[robot_state_publisher-3] Link base had 0 children
[robot_state_publisher-3] Link base_link_inertia had 1 children
[robot_state_publisher-3] Link shoulder_link had 1 children
[robot_state_publisher-3] Link upper_arm_link had 1 children
[robot_state_publisher-3] Link forearm_link had 1 children
[robot_state_publisher-3] Link wrist_1_link had 1 children
[robot_state_publisher-3] Link wrist_2_link had 1 children
[robot_state_publisher-3] Link wrist_3_link had 1 children
[robot_state_publisher-3] Link flange had 1 children
[robot_state_publisher-3] Link tool0 had 0 children
[...]

PR #266, opened 2021-12-16
Do not print an error output if writing is not possible

fmauch:
If writing to the hardware is not possible, as the interpreting program
is currently not running, there's no need to print an error message in each
control cycle.

This aims at fixing #262 and possibly #264

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -533,7 +533,7 @@ hardware_interface::return_type URPositionHardwareInterface::read()
 hardware_interface::return_type URPositionHardwareInterface::write()
 {
   // If there is no interpreting program running on the robot, we do not want to send anything.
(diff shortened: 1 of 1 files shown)


ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -532,6 +532,9 @@ hardware_interface::return_type URPositionHardwareInterface::read()
 
 hardware_interface::return_type URPositionHardwareInterface::write()
 {
+  // If there is no interpreting program running on the robot, we do not want to send anything.
+  // TODO: We would still like to disable the controllers requiring a writable interface. In ROS1
```
destogl, line 536: I think linter give here error because it is expected following line: 
```suggestion
  // TODO(anyone): We would still like to disable the controllers requiring a writable interface. In ROS1
```

destogl (approved):
All good, as discussed.

livanov93 (approved):

Issue #256, opened 2021-12-09
Test tool communication


livanov93:
@fmauch After discussion could you please give us some feedback on testing tool communication after #218 has been merged.
Thanks!

fmauch:
yep, I'll do that.

fmauch:
I wasn't able to test the full pipeline on a real robot as I couldn't find the connection cable for my hardware in order to connect it to the UR. The script itself should work though, as it is the same that is used for the ROS1 driver.

However, I noticed that the tool_communication script does not get started automatically, when `use_tool_communication` is set to `true` in the launchfile. For the ROS1 driver we actually [started the script](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/launch/ur_control.launch#L61) when that argument was set.

livanov93:
Sounds good to start the script if the parameter is set to `true`. I will address this in a follow-up PR

PR #267, opened 2021-12-16
Start the tool communication script if the flag is set

fmauch:
As discussed in #256 it would be nice to actually start the tool communication if the flag `use_tool_communication` is set.

A quick side note: I decided against adding this to all launch files, as the `use_tool_communication` argument also only exists inside the `ur_control.launch` file.

I'll mark this ready once I've actually tested on real hardware.

### DIFF
--- a/ur_bringup/launch/ur_control.launch.py
+++ b/ur_bringup/launch/ur_control.launch.py
@@ -45,6 +45,14 @@ def launch_setup(context, *args, **kwargs):
     headless_mode = LaunchConfiguration("headless_mode")
     launch_dashboard_client = LaunchConfiguration("launch_dashboard_client")
     use_tool_communication = LaunchConfiguration("use_tool_communication")
+    tool_parity = LaunchConfiguration("tool_parity")
+    tool_baud_rate = LaunchConfiguration("tool_baud_rate")
+    tool_stop_bits = LaunchConfiguration("tool_stop_bits")
+    tool_rx_idle_chars = LaunchConfiguration("tool_rx_idle_chars")
+    tool_tx_idle_chars = LaunchConfiguration("tool_tx_idle_chars")
+    tool_device_name = LaunchConfiguration("tool_device_name")
+    tool_tcp_port = LaunchConfiguration("tool_tcp_port")
+    tool_voltage = LaunchConfiguration("tool_voltage")
 
     joint_limit_params = PathJoinSubstitution(
         [FindPackageShare(description_package), "config", ur_type, "joint_limits.yaml"]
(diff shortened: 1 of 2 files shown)


ur_bringup/launch/ur_control.launch.py
```diff
@@ -45,6 +45,13 @@ def launch_setup(context, *args, **kwargs):
     headless_mode = LaunchConfiguration("headless_mode")
     launch_dashboard_client = LaunchConfiguration("launch_dashboard_client")
     use_tool_communication = LaunchConfiguration("use_tool_communication")
+    tool_parity = LaunchConfiguration("tool_parity")
+    tool_baud_rate = LaunchConfiguration("tool_baut_rate")
```
gavanderhoorn, line 49:
```suggestion
    tool_baud_rate = LaunchConfiguration("tool_baud_rate")
```

fmauch:
Tested with real hardware.

I found an issue with the tool_comm script which I also fixed in this PR.

ur_bringup/launch/ur_control.launch.py
```diff
@@ -125,6 +133,30 @@ def launch_setup(context, *args, **kwargs):
...
+            " ",
+            "tool_device_name:=",
+            tool_device_name,
+            " ",
+            "tool_tcp_port:=",
+            tool_tcp_port,
+            " ",
+            "tool_voltage:=",
+            tool_voltage,
+            " ",
```
destogl, line 159: The parameters here are leading to "nirvana" (read nowhere) :smiley: 

You would probably like to add them into `ur.urdf.xacro`, `ur_macro.xacro` and `ur.ros2_control.xacro`.

Or I am missing something here?

I think that your example works because you are using TCP tool, but if you start to use RS485 tool, this would not be configured.

fmauch: Aren't they read [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/a0cfee50ae7b9318ad705d9634716a76cdb131a7/ur_description/urdf/ur.ros2_control.xacro?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L35-L42)?

destogl: Yes, sorry. All good. I was looking to the wrong file.

destogl (approved):
Looking good!

PR #268, opened 2021-12-22
Fixing launch files ⛏: Fix typo and changes when launching newest MoveIt on rolling.

destogl:
Update wrong description for “dashboard” parameter.

MoveIt now uses proper node parameters, so parameter substitution is possible/needed instead of manual YAML parsing.﻿ Without this change, I get the following error because parameter type is interpreted wrongly.

moveit_error_wrong_parameter_type

### DIFF
--- a/ur_bringup/launch/ur_control.launch.py
+++ b/ur_bringup/launch/ur_control.launch.py
@@ -368,7 +368,7 @@ def generate_launch_description():
     )
     declared_arguments.append(
         DeclareLaunchArgument(
-            "launch_dashboard_client", default_value="true", description="Launch RViz?"
+            "launch_dashboard_client", default_value="true", description="Launch Dashboard client?"
         )
     )
     declared_arguments.append(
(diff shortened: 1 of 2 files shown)


antopius:
Thanks for the commit!
I am using Galactic and with #268 the Robotmodel loads, but after that i get the [Info]
'No Kinematics. Please load kinematics.yaml'

Had someone experienced the same on Galactic or is there no relation to this commit?

destogl:
> Thanks for the commit! I am using Galactic and with #268 the Robotmodel loads, but after that i get the [Info] 'No Kinematics. Please load kinematics.yaml'
> 
> Had someone experienced the same on Galactic or is there no relation to this commit?

Does main works for you?

antopius:
> Does main work for you?

Unfortunately, no. The same error occurred as you mentioned above.
I will open a new issue tomorrow with some more informations, as this could likewise related to something else.

Thanks for you Support!

destogl:
PR is replaced by #278. The issue has to be investigated more. This is rather a MoveIt2 issue with loading parameters.

PR #272, opened 2022-01-04
Backport ursim_driver to foxy branch

### LINKED ISSUE #270: ursim_driver missing on the foxy branch
Does it need to be back ported, perhaps?

destogl:
resolves #270


urrsk:
In stead of building the ursim docker, we could use the image on docker hub.
See:
- https://hub.docker.com/repository/docker/universalrobots/ursim_e-series
- https://hub.docker.com/repository/docker/universalrobots/ursim_cb3

livanov93 (approved):
It looks good to me. I am not sure about @urrsk comment, should we use simplify the ursim dockerizing by pulling it from dockerhub.
I think it is fine like this now. And when we do the simplification, we can do another backport.
@destogl @urrsk Do you agree?

ur_robot_driver/resources/ursim_driver/driver/Dockerfile
```diff
@@ -0,0 +1,86 @@
+# -- BEGIN LICENSE BLOCK ----------------------------------------------
```
urrsk, line 1: Please change it to BSD3

ur_robot_driver/resources/ursim_driver/ursim/aux/ros_control.urscript
```diff
@@ -0,0 +1,143 @@
+{{BEGIN_REPLACE}}
```
urrsk, line 1: Why is this file needed. It should use the one from the client_library!

ur_robot_driver/resources/ursim_driver/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+   ports:
+      - "5900:5900"
+      - "29999:29999"
+      - "8080:8080"
+      - "30001-30004:30001-30004"
+   volumes:
+      - ./ursim/aux/.vol/:/ursim/programs
+   build:
+    context: ursim/.
+    dockerfile: g5/Dockerfile
```
fmauch, line 41: We could use the upstream docker images released by UR by now. But that's not a must.

ur_robot_driver/resources/ursim_driver/ursim/aux/.vol/urcap_ros_control.script
```diff
@@ -0,0 +1,180 @@
+def urcap_ros_control():
```
fmauch, line 1: Why is this moved to the robot already? This should come from our driver, right?

ur_robot_driver/resources/ursim_driver/ursim/aux/ros_control.urscript
fmauch: agreed

ur_robot_driver/resources/ursim_driver/ursim/g3/Dockerfile
```diff
@@ -0,0 +1,112 @@
+# MIT License
```
fmauch, line 1: I think, using the released docker images would be better.

fmauch (commented):
Over all, I think, it would be better to use the released upstream images. In a tutorial package I recently started, I created a rather slim version of an example bringing up preconfigured robots with the urcap installed and a program loaded: https://github.com/UniversalRobots/Universal_Robots_ROS_Tutorials/blob/main/ur_example_dual_robot/resources/docker-compose.yml

However, as this is already the state that we have on the main branch, I would also be OK with merging it like this.

fmauch:
I'll remove this from the Foxy release milestone, as I am still not sure whether we want this in like that. I'd like to change the tests a bit more using the official URSim Dockerfile and I am not sure how much sense it makes to have a docker-compose file in this repo actually building and running the driver...

fmauch:
This will not be done anymore for Foxy.

Issue #265, opened 2021-12-16
a lot Parameters have invalid type: is of type {double}, setting it to {string} is not allowed


NebuDa:
I have built ROS2 Galactic, Moveit2 and Main Branch of this driver on Ubuntu. By starting like the tutroial in the readme with MoveIt I got the following error on different parameters:
```
[rviz2-3] [INFO] [1639652465.916282193] [moveit_rdf_loader.rdf_loader]: Loaded robot model in 0.00852234 seconds
[rviz2-3] [INFO] [1639652465.917162913] [moveit_robot_model.robot_model]: Loading robot model 'ur'...
[rviz2-3] [ERROR] [1639652465.936346953] [moveit_background_processing.background_processing]: Exception caught while processing action 'loadRobotModel': parameter 'robot_description_planning.joint_limits.shoulder_pan_joint.max_position' has invalid type: Wrong parameter type, parameter {robot_description_planning.joint_limits.shoulder_pan_joint.max_position} is of type {double}, setting it to {string} is not allowed.
```
Another parameter is by example: robot_description_kinematics.ur_manipulator.kinematics_solver_timeout

Is there anything wrong with Ubuntu or ROS2 configuration, that it tries to load it as string?

P.S.
Perhaps it could have something to do with it:
I am not allowed now to publish a double with to_string to the RCLCPP Logger. It was absolutely working under Foxy on the same Machine.

`RCLCPP_INFO(LOGGER, std::to_string(q.x()));`

destogl:
I am guessing that you are using rolling. In any case, please check #268 if it solves your problem and comment there.

fmauch:
This should be resolved by now.

PR #278, opened 2022-01-21
Update MoveIt file for working with simulation.

### LINKED ISSUE #261: It feels weird to provide driver parameters to the moveit launchfile
First of all: Great, that there is a nice MoveIt integration out of the box!

I was a bit puzzled from the fact that I had to specify the robot's IP address to the moveit launchfile and double-checked that it did not run another driver instance internally.

I understand the rationale behind this, as hardware parameters are passed through the `ur.ros2_control.xacro` file and in ROS2 every node needing the description has to get it individually, but it feels wrong nevertheless, that all of those parameters are exposed inside the moveit launchfile, while the `robot_ip` is even required.

From quickly scanning over it I would assume that the `robot_ip` and `headless_mode` arguments should be removed from the launchfile as passing them doesn't make any sense and might even confuse new users.

Please feel free to correct me if I am wrong with my assumptions above, I just try to note down every confusing bit that I stumbled upon, being an experienced ROS1 developer/user, but relatively new to ROS2.

destogl:
Update of MoveIt launch files to be a bit shorter/simpler and to working with simulation.﻿

should resolve #261


ur_moveit_config/package.xml
```diff
@@ -11,6 +11,16 @@
   <license>Apache2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>ament_cmake_python</buildtool_depend>
+
+  <exec_depend>launch</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
+  <exec_depend>moveit_ros_move_group</exec_depend>
+  <exec_depend>moveit_servo</exec_depend>
```
destogl, line 19:
```suggestion
```

ur_moveit_config/launch/ur_moveit.launch.py
```diff
@@ -147,12 +123,15 @@ def launch_setup(context, *args, **kwargs):
...
+        [FindPackageShare(moveit_config_package), "config", "kinematics.yaml"]
+    )
     robot_description_kinematics = {"robot_description_kinematics": kinematics_yaml}
 
-    robot_description_planning = {
-        "robot_description_planning": load_yaml_abs(str(joint_limit_params.perform(context)))
-    }
+    # robot_description_planning = {
+    # "robot_description_planning": load_yaml_abs(str(joint_limit_params.perform(context)))
+    # }
```
fmauch, line 134: Why has this been commented out? That seems to remove functionality.

fmauch (commented):
Apart from the removed `robot_description_planning` this looks good to me. However, I cannot estimate how important that is, so feel free to ignore this comment.

ur_moveit_config/launch/ur_moveit.launch.py
destogl: This currently make issues when loaded. The parameters can not be parsed properly.
It is related to #265 and ros-planning/moveit2_tutorials#303.

Without it everything works, but to have things fully configurable we'll have to wait for upstream fixes.

fmauch (dismissed):
lgtm. Thanks for the clarifications.

fmauch:
Sorry, for withdrawing my approval... I noticed, that the README file should also be adapted, since the moveit launchfile was moved around.

destogl:
> Should we address this here or in a separate PR afterwards? As there are also a couple of other dependencies missing for example in `ur_bringup` we might want to address this in a separate PR, actually.

let's fix all the dependencies in a separate PR.

destogl:
@fmauch it seems that everything is resolved here. Can we proceed with merging? This is causing some issues on other repos like UniversalRobots/Universal_Robots_ROS2_Gazebo_Simulation#8

fmauch (approved):
As we want to fix the dependencies in a separate PR, this should be fine as it is.

PR #281, opened 2022-01-27
Separate control node

livanov93:
PR info:
Adds separate ros2 control node without sleeping in the main loop as a way to solve the [`Pipeline produced overflow!` problem](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/210).

Instead of #263.


destogl (approved):
Looking good! Thanks!

urrsk (approved):
It looks good, but there are some failures in the CI.
Please consider fixing them before merging

PR #287, opened 2022-02-02
Use scaled trajectory controller per default.

### LINKED ISSUE #260: Why is the non-scaled trajectory controller the default controller?
If not changed manually, the default controller [is the joint_trajectory_controller](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/a22e5bc586a611866918473eafc04232111ab991/ur_bringup/launch/ur5e.launch.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L50). Since the driver supports using the scaled version of that controller out of the box it might make sense to use that as default instead, so users can make direct use of the speed scaling feature.

Is there any incentive behind not specifying the scaled version by default or has this simple been missed?

destogl:
resolves #260﻿

### DIFF
--- a/ur_bringup/launch/ur10.launch.py
+++ b/ur_bringup/launch/ur10.launch.py
@@ -47,7 +47,7 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument(
             "initial_joint_controller",
-            default_value="joint_trajectory_controller",
+            default_value="scaled_joint_trajectory_controller",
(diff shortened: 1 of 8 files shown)


fmauch (approved):
lgtm, thanks!

Issue #296, opened 2022-02-18
Fresh galactic install fails to run examples


maffettone:
When trying to run the initial examples with ros2 galactic, I get the following error. 

I removed all other ros installations `sudo apt remove ~nros-foxy-* && sudo apt autoremove` and `sudo apt remove ~nros-galactic-* && sudo apt autoremove`, and started from scratch installing only galactic on the system. Using Ubuntu 20.04.3 LTS. 

```bash
ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=true
[INFO] [launch]: All log files can be found below /home/pmm/.ros/log/2022-02-17-20-09-30-567585-ROS-Workstation-83035
[INFO] [launch]: Default logging verbosity is set to INFO
Task exception was never retrieved
future: <Task finished name='Task-2' coro=<LaunchService._process_one_event() done, defined at /opt/ros/galactic/lib/python3.8/site-packages/launch/launch_service.py:226> exception=SubstitutionFailure("executable 'spawner.py' not found on the libexec directory '/home/pmm/Development/ros_ur_driver/install/controller_manager/lib/controller_manager' ")>
Traceback (most recent call last):
  File "/opt/ros/galactic/lib/python3.8/site-packages/launch/launch_service.py", line 228, in _process_one_event
    await self.__process_event(next_event)
  File "/opt/ros/galactic/lib/python3.8/site-packages/launch/launch_service.py", line 248, in __process_event
    visit_all_entities_and_collect_futures(entity, self.__context))
  File "/opt/ros/galactic/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
    futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
  File "/opt/ros/galactic/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
    futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
  File "/opt/ros/galactic/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 45, in visit_all_entities_and_collect_futures
    futures_to_return += visit_all_entities_and_collect_futures(sub_entity, context)
[...]

PR #297, opened 2022-02-18
Update readme install instructions

fmauch:
As risen for example in #296 our current installation instructions build a lot from source using the default branches. Obviously, this introduces changes not really applicable to older ROS versions.

For example, ros2_control dropped support for the deprecated `spawner.py` on the main branch. Therefore, I suggest changing the README to use the distro-specific dependency definitions and I updated the galactic definitions here.

In the long run we should try to reduce those as well, as packages from our `not-released...` rosdistro files are indeed released. But that's for another PR...

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ ROS2 Distro | Foxy  | Galactic | Rolling
    ```
    cd $COLCON_WS
    git clone https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver.git src/Universal_Robots_ROS2_Driver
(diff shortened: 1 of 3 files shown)


destogl:
> In the long run we should try to reduce those as well, as packages from our not-released... rosdistro files are indeed released. But that's for another PR...

Fully agree. Can you contact me directly at some point to discuss shortly release strategy, i.e., what is missing from ros2_control to get packages released?

destogl (approved):

fmauch:
As far as I have found so far, there are no other ros2_control packages missing.

PR #300, opened 2022-02-23
Make documentation on how to use driver clearer.

destogl:


README.md
```diff
@@ -176,32 +176,48 @@ The most relevant arguments are the following:
...
+  ```
+
+  **NOTE**: Instead of using global launch file for control stack, there are also prepeared launch files for each type of UR robots named. The accept the same arguments are the global one and are used by:
+  ```
+  ros2 launch ur_bringup <ur_type>.launch.py
+  ```
+
+##### 2. Sending commands to controllers
+
+**NOTE**: If controllers are not starting automatically, i.e., the robot state is not shown in rviz, you can start them manually:
```
fmauch, line 205: If they do not get started automatically, I would assume that there is something severely wrong, right? So, trying to start the `joint_state_broadcaster` in that case would probably only lead to more error output.

fmauch (commented):
Looking good apart from one spelling thing.

README.md
destogl: Probably yes. This was an old solution to an old issue. Thinking more about it, you are right. This part is obsolete...

README.md
```diff
@@ -176,32 +176,48 @@ The most relevant arguments are the following:
...
+```
+ros2 control load_controller --set-state start joint_state_broadcaster
+ros2 control load_controller --set-state start scaled_joint_trajectory_controller
+```
+To check the controllers' state use `ros2 control list_controllers` command.
```
destogl, line 210:
```suggestion
Before running any commands, first check the controllers' state using `ros2 control list_controllers`.
```

fmauch (approved):
Looks good to me

PR #304, opened 2022-02-23
Use upstream fts_broadcaster

fmauch:
As already proposed in the README this uses the upstream FTS broadcaster


destogl (approved):
This looks good and interfaces are correct/compatible so deleting controller is sufficient :)

PR #306, opened 2022-02-24
Added missing exec_dependencies

### LINKED ISSUE #292: ros2_control_test_nodes is a missing exec_depend but not available
Currently, we have launchfiles using `ros2_control_test_nodes` which are also referenced in our README.md.

I'd create a PR adding the dependency, but as it is not released for ROS galactic we should make our heads clear whether we want to keep it. Once we want to do a binary release we need to have all dependencies available in released form.

@destogl what are your assumptions on the chances of having this released?

If we don't see that `ros2_control_test_nodes` will be released, we might want to get rid of the launchfiles before we release the driver.

Currently, this affects
 * [ur_bringup/launch/test_forward_velocity_controller.launch.py](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_bringup/launch/test_forward_velocity_controller.launch.py)
 * [ur_bringup/launch/test_joint_trajectory_controller.launch.py](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_bringup/launch/test_joint_trajectory_controller.launch.py)
 * [ur_bringup/launch/test_scaled_joint_trajectory_controller.launch.py](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_bringup/launch/test_scaled_joint_trajectory_controller.launch.py)

fmauch:
This resolves #292 

As `ros2_controllers_test_nodes` has been released into latest `ros2_controllers` I'll suggest waiting to merge this until the pipeline actually gets green.


destogl (approved):

fmauch:
I think this PR should strive towards making all pipelines work again. So, let's wait with a final judgement

fmauch:
With further investigation, it seems that Jammy's transition to python 3.10 seems to be causing trouble (see also [here](https://discourse.ros.org/t/preparing-for-rolling-sync-delayed-by-python-3-10-transition/24521)). In that case I might want to vote for merging this beforehand. We can fix the rolling build later if it turns out that there was something missing.

fmauch:
I'll go ahead and merge this as this is currently blocking me from other MRs such as #309

PR #308, opened 2022-02-25
Load kinematics parameters as yaml

### LINKED ISSUE #307: Galactic: the “Interact” tool does not work for “ur_manipulator” when following the moveit example
The “Interact” tool  seems not working when trying to set the goal for the “ur_manipulator”.  The log which seems relevant to that is:

`[rviz2-3] [INFO] [1645710748.640870337] [moveit_ros_robot_interaction.robot_interaction]: No active joints or end effectors found for group 'ur_manipulator'. Make sure that kinematics.yaml is loaded in this node's namespace.
[rviz2-3] [INFO] [1645710748.646816651] [moveit_ros_visualization.motion_planning_frame]: group ur_manipulator
[rviz2-3] [INFO] [1645710748.646855345] [moveit_ros_visualization.motion_planning_frame]: Constructing new MoveGroup connection for group 'ur_manipulator' in namespace ''`

fmauch:
This fixes #307 which outlines a regression introduced in #278

~~@destogl As we use the raw string "ur_moveit_config" in several other places, it probably doesn't harm to leave this.~~
With the implementation from bf37c75 it should be possible to keep both.

### DIFF
--- a/ur_moveit_config/config/kinematics.yaml
+++ b/ur_moveit_config/config/kinematics.yaml
@@ -1,5 +1,8 @@
-ur_manipulator:
-  kinematics_solver: kdl_kinematics_plugin/KDLKinematicsPlugin
-  kinematics_solver_search_resolution: 0.005
-  kinematics_solver_timeout: 0.005
-  kinematics_solver_attempts: 3
+/**:
+  ros__parameters:
+    robot_description_kinematics:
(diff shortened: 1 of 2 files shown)


destogl (approved):
This solution is very, very, very welcome!!! Thanks for finding this out!

PR #311, opened 2022-02-28
Ci do not fail fast

fmauch:
Do not fail fast in CI runs. We'd like to know if it is just  a flaw in the testing repos or if there is something wrong with main.

### DIFF
--- a/.github/workflows/galactic-binary-build.yml
+++ b/.github/workflows/galactic-binary-build.yml
@@ -15,6 +15,7 @@ jobs:
     name: galactic binary build
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         env:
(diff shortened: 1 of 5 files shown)


fmauch:
I had this hanging in Draft for a while as the Rolling build was still failing with this due to other reasons. However, I think it is worth merging this, as this will help debugging why the other builds as for example the semi-binary galactic build fail.

fmauch:
I actually remembered why I had this in WIP: I wanted to migrate the workflows to a github-compatible format.

destogl (approved):
Looks good!

P.S. I am currently working even more on CI simplification. Will let you know if my concept brings some fruits. (Basically, I am attempting to reduce the amount of redundant code we currently have all around the place. — not only this repository, but other ROS repositories as well.)

PR #312, opened 2022-03-03
Fix scheduled foxy build

fmauch:
The scheduled build is running on the default branch if not told differently.


fmauch:
Reopened to trigger the CI pipeline

t-schnell (approved):
As foxy seems to build fine, everything looks good.

PR #319, opened 2022-03-25
Re-add foxy builds to main branch

fmauch:
Obviously I was not aware that workflow definitions have to reside on the main branch...

### DIFF
--- a/.github/workflows/foxy-binary-build.yml
+++ b/.github/workflows/foxy-binary-build.yml
@@ -1,8 +1,5 @@
 name: Foxy Binary Build
 on:
-  pull_request:
-    branches:
-      - foxy
   push:
     branches:
       - foxy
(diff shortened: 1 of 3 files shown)


fmauch:
Actually, I am not sure whether this is correct...

.github/workflows/foxy-binary-build.yml
```diff
@@ -0,0 +1,61 @@
+name: Foxy Binary Build
+on:
+  push:
+    branches:
```
t-schnell, line 4: Maybe remove this here and only leave it on the foxy branch - similar to the galactic solution?

.github/workflows/foxy-semi-binary-build.yml
```diff
@@ -0,0 +1,61 @@
+name: Foxy Semi-Binary Build
+on:
+  push:
+    branches:
```
t-schnell, line 4: as above

.github/workflows/foxy-source-build.yml
```diff
@@ -0,0 +1,45 @@
+name: Foxy Source Build
+on:
+  push:
+    branches:
```
t-schnell, line 4: as above

t-schnell (approved):
As written at the point in the review - we maybe should remove the on push action of the branch in main, and only leave the scheduled actions to be consistent to the way it is handled with galactic (unless I'm missing something). As this is not critical, I will approve either way.

PR #309, opened 2022-02-25
Added controller stopper node

urmahp:
* This will stop and start controllers sending commands to the robot through the hardware interface.

* Added programm_state publisher in the gpio controller

* This should fix issue #279

PR #324, opened 2022-04-01
Foxy controller stopper

fmauch:
Backport of #309 

This is currently a draft, as I haven't tested it, yet.

### DIFF
--- a/ur_controllers/src/gpio_controller.cpp
+++ b/ur_controllers/src/gpio_controller.cpp
@@ -101,6 +101,9 @@ controller_interface::InterfaceConfiguration ur_controllers::GPIOController::sta
     config.names.emplace_back("gpio/safety_status_bit_" + std::to_string(i));
   }
 
+  // program running
+  config.names.emplace_back("gpio/program_running");
+
   return config;
 }
 
@@ -117,6 +120,7 @@ controller_interface::return_type ur_controllers::GPIOController::update()
   publishToolData();
   publishRobotMode();
   publishSafetyMode();
+  publishProgramRunning();
   return controller_interface::return_type::OK;
 }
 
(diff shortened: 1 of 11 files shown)


fmauch:
@urmahp would you like helping me out in testing this? This currently compiles fine, but running it crashes.

urmahp:
> @urmahp would you like helping me out in testing this? This currently compiles fine, but running it crashes on this line:
> 
> https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/b2014e060493509b6a9a848dfd30d0c55670da42/ur_controllers/src/gpio_controller.cpp#L203
> 
> Seems like `state_interfaces_[StateInterfaces::PROGRAM_RUNNING]` is not returning a valid object. I am not entirely sure why this happens, though.

I found the issue, the intialized flag is not added to the state interfaces in the hardware interface on the foxy branch, however it is on the main branch, see [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/src/hardware_interface.cpp#L213-L214). This means that the when we try to index program_running the index is out of range and therefore the program crashes. 

I see two possible fixes adding the initialized flag and using it during controller startup (which i believe is where it is used) or simply removing it which will also solve the problem. However I prefer the first one. I can fix it if you agree?

fmauch:
> I see two possible fixes adding the initialized flag and using it during controller startup (which i believe is where it is used) or simply removing it which will also solve the problem. However I prefer the first one. I can fix it if you agree?

I agree, the first one would be preferable. Thank you @urmahp

fmauch:
Aweseome, thanks @urmahp I'll update this PR to be reviewable and then we can proceed with this.

ur_controllers/src/gpio_controller.cpp
```diff
@@ -100,6 +100,10 @@ controller_interface::InterfaceConfiguration ur_controllers::GPIOController::sta
   for (size_t i = 0; i < 11; ++i) {
     config.names.emplace_back("gpio/safety_status_bit_" + std::to_string(i));
   }
+  config.names.emplace_back("system_interface/initialized");
```
fmauch, line 103: That would probably require adding this to the xacro, as well. (For Foxy we decided to stick with the one in this repo)

fmauch: resolved in [7ea0d27](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/324/commits/7ea0d27263aa4bcba4f059a3db4582e1221ffe71)

PR #329, opened 2022-04-07
Fix hardware interface names in error output

### LINKED ISSUE #328: Incorrect state interface name in error message
The error messages for having an incorrect state interface for all state interfaces say that position is expected for the state interfaces. The velocity and effort state interfaces should probably say that velocity/effort is expected instead of position.

https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/ff3517dff7bd69fafd755b07b6fecc31b0a3ebc5/ur_robot_driver/src/hardware_interface.cpp#L109-L120

fmauch:
They were all referring to hardware_interface::HW_IF_POSITION also for
velocity and effort interfaces

Fixes #328


t-schnell (approved):
No comments, as this is a simple and obvious fix.

Issue #323, opened 2022-04-01
Update README regarding MoveIt!


fmauch:
We should change the MoveIt section in the README as it is currently suggesting to build MoveIt from source. This is, however, only necessary for Rolling. Building it from source with the given repo file might actually lead to problems on distributions different from rolling (see #322 for example).

I'd suggest removing the install part completely and instead add MoveIt to the `rolling.repos` file.

fmauch:
Additionally: We could add `rosdep update` to the installation instructions.

fmauch:
Addressed in #333

PR #333, opened 2022-04-07
Update README regarding MoveIt setup

fmauch:
As proposed in #323 we should suggest using the binary-installed MoveIt if applicable. I changed the README accordingly and moved the MoveIt repos to the rolling repos files instead.

As we currently require changing the default controller for MoveIt when using fake_hardware, I also updated the README accordingly.


fmauch:
@destogl I am rather unhappy with the default_controller issue regarding the fake_hardware. Do you have a better idea how we could handle that?

The core problem there is, that we don't have the `scaled_jtc` available for fake_hardware.

destogl:
> The core problem there is, that we don't have the scaled_jtc available for fake_hardware.

Is this also an issue with `fake_hardware`? I thought that are only simulators. `fake_hardware` should support that, but maybe some minor parameter changes are needed.

fmauch:
> 

Thanks for letting me know, I'll take another look into that.

fmauch:
> Is this also an issue with `fake_hardware`? I thought that are only simulators. `fake_hardware` should support that, but maybe some minor parameter changes are needed.

It seems that the `speed_scaling` value is `NaN` for fake_hardware. This makes the controller crash. I guess, this should be fixed and therefore issued at the description?

README.md
```diff
@@ -66,22 +66,74 @@ ROS2 Distro | Foxy  | Galactic | Rolling
...
-To use MoveIt some additional packages should be added into workspace:
-   ```
-   cd $COLCON_WS
-   vcs import src --skip-existing --input src/Universal_Robots_ROS2_Driver/MoveIt_Support.repos
-   vcs import src --skip-existing --input src/moveit2/moveit2.repos
-   rosdep install --ignore-src --from-paths src -y -r
-   colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
-   source install/setup.bash
-   ```
+MoveIt! support is built-in into this driver already.
```
t-schnell, line 77: Maybe add a link to MoveIt! page/installation here?

README.md
```diff
@@ -66,22 +66,74 @@ ROS2 Distro | Foxy  | Galactic | Rolling
...
+
+### Real robot / URSim
+To test the driver with the example MoveIt-setup, first start the driver as described
+[below](#connect-to-exnernal-control-via-urcap).
```
t-schnell, line 81:
```suggestion
[below](#connect-to-external-control-via-urcap).
```

README.md
```diff
@@ -66,22 +66,74 @@ ROS2 Distro | Foxy  | Galactic | Rolling
...
+    - wrist_2_joint
+    - wrist_3_joint
+```
+
+Then start
+
+```
+ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=false initial_joint_controller:=joint_trajectory_controller
+# and in another shell
+ros2 launch ur_moveit_config ur_moveit.launch.py ur_type:=ur5e launch_rviz:=true
```
t-schnell, line 133: I assume this is intended for people who know MoveIt! and just need information about how to set up the fake hardware right? If so, this should be fine, but otherwise I guess there should be some more information or a link to MoveIt instructions.

t-schnell (approved):
Minor optional comments and a typo, no further review should be necessary.

README.md
fmauch: Yep, that was meant only as a guide how to get it running. But a link to the according MoveIt! tutorial might definitively be useful.

PR #338, opened 2022-04-12
Fix precommit black foxy

fmauch:
Backport of #321


fmauch:
It happens all the time, that I start making foxy PRs on the main branch... sorry for that.

PR #339, opened 2022-04-14
Use types in hardware interface from ros2_control in local namespace

destogl:
ros2_control moved some of the global usings to the local namespace. The change is compatible with galactic and rolling. It correct builds from the source on rolling but does not break anything.

ros-controls/ros2_control#673


fmauch (dismissed):
Makes sense and shouldn't do any harm.

fmauch:
But wait a second. Should this be necessary at all?

We include `hardware_interface/system_interface.hpp`, which defines `hardware_interface::CallbackReturn`. Wouldn't it be better to reuse that instead of redefining it ourselves? As er are implementing the virtual methods from `hardware_interface::SystemInterface` it would only make sense to use its return type definition.

destogl:
> We include `hardware_interface/system_interface.hpp`, which defines `hardware_interface::CallbackReturn`. Wouldn't it be better to reuse that instead of redefining it ourselves? As we are implementing the virtual methods from `hardware_interface::SystemInterface` it would only make sense to use its return type definition.

You are right! This is much better! Thanks for the idea!!!

destogl:
@fmauch updated! This may be failing on galactic though.

fmauch:
Thanks @destogl. That might be the time to actually branch out for galactic...

Let's finish a couple of outstanding PRs, then branch out and merge this.

fmauch (approved):
rebased, still aprove

PR #340, opened 2022-04-19
Add sphinx documentation

fmauch:
Migrate the `docs` folder towards using sphinx. My goal here is to have a complete documentation that can be built using [`rosdoc2`](https://github.com/ros-infrastructure/rosdoc2) so we have a place for a rendered documentation outside the repo once we use binary releases.


fmauch:
This will currently conflict with #333 but that shouldn't be a problem to update once #333 is merged.

destogl (approved):
This looks fine to me.

DISCLAIMER: I only checked the files and didn't try to build this locally, because I am not sure how to use `rosdoc2`

PR #285, opened 2022-02-01
Update header extension to remove compile warning.

destogl:
This change removes compile warning on Rolling:

```
In file included from /home/deniss/workspace/robosoft-ai/smacc_rolling/src/Universal_Robots_ROS2_Driver/ur_robot_driver/include/ur_robot_driver/hardware_interface.hpp:49,
                 from /home/deniss/workspace/robosoft-ai/smacc_rolling/src/Universal_Robots_ROS2_Driver/ur_robot_driver/src/hardware_interface.cpp:36:
/opt/ros/rolling/include/tf2_geometry_msgs/tf2_geometry_msgs.h:35:2: warning: #warning This header is obsolete, please include tf2_geometry_msgs/tf2_geometry_msgs.hpp instead [-Wcpp]
   35 | #warning This header is obsolete, please include tf2_geometry_msgs/tf2_geometry_msgs.hpp instead
```

PR #345, opened 2022-04-19
Fix coverage build

fmauch:
as per subject...


fmauch:
This actually does not make the build succeed due to #285, so I would be ok with marking this `humble+`.

t-schnell (approved):

fmauch:
Merging this as this will be the base for other fixes

fmauch:
Merging this as this will be the base for other fixes

PR #351, opened 2022-04-28
Update controllers' API

destogl:
Update ur_controller to new interface introduced in:

- ros-controls/ros2_control#699
- ros-controls/ros2_control#701


fmauch (approved):
LGTM, but is humble+

PR #360, opened 2022-05-02
Updated package maintainers

fmauch:
Corrected tag order according to schema while at it.

@destogl I assume you'd like to keep being listed as a maintainer. Which mail address should I use for that?


t-schnell (approved):

fmauch:
Discussed offline with @destogl to update his mail address. Therefore merging this.

PR #362, opened 2022-05-03
Replace warehouse_ros_mongo with warehouse_ros_sqlite

fmauch:
Since mongodb is not available on Ubuntu jammy, I don't see `warehouse_ros_mongodb` is coming on Jammy any time soon. See upstream https://github.com/ros-planning/warehouse_ros_mongo/issues/71 for details.

Keeping this draft as I currently don't have a jammy system to test this on.


fmauch:
Once other issues with the rolling build have been resolved (PR for those coming soon), this should be building correctly.

PR #363, opened 2022-05-03
Remove warehouse plugin for MoveIt (for now)

fmauch:
As no plugin is currently available in ROS2 Rolling (on Ubuntu Jammy), let's
remove support for it for now, so we can get successful builds again.

This is an alternative to #362 as the sqlite-based warehouse also seems not to build currently.

As far as I can see, removing the warehouse should only disable persistent storage functionality but leave the rest working.


fmauch:
Closing this in favor of #362

PR #366, opened 2022-05-05
Add dependency to ur_robot_driver

fmauch:
Otherwise installing `ros-galactic-ur-bringup` won't install a functioning thing...


t-schnell (approved):

fmauch:
Actually, I think I'd like to discuss this a little more... I think the following questions would have to be answered:
 - Do we want users to install `ur_bringup` or `ur_robot_driver` as "main" package?
 - What would be the "expected" dependency order?
 - How do we solve #367?

As for now I would be fine with leaving things as is, as users can install `ros-galactic-ur-robot-driver` and everything should work for them.

fmauch:
This is superseeded by #403

PR #385, opened 2022-06-07
Add galactic source build on galactic branch

fmauch:
This will enable source builds on pushes to the galactic branch


RobertWilbrandt (changes requested):
Please change the name to ```galactic-source-build.yml``` to be consistent with the other workflows, otherwise LGTM

PR #386, opened 2022-06-07
Remove source builds

fmauch:
The way github actions and the ros-ci action are implemented, this will not work as desired.

For GH actions the schedule has to run on the default branch, so the git ref will be the one from the default branch. As the ros-ci action forces the repo to use the version specified in the GIT_HEAD_REF (see https://github.com/ros-tooling/action-ros-ci/blob/4cc03e8318d61efb18eb08835e07f70df1b9cbf6/src/action-ros-ci.ts#L519-L533) this will always run against the default branch.


RobertWilbrandt (approved):
LGTM

PR #393, opened 2022-06-13
Force galactic branch for scheduled builds

fmauch:
As scheduled runs have to run on the main branch, we have to explicitly
checkout the distro's branch in the CI run.


RobertWilbrandt (approved):
LGTM

PR #396, opened 2022-06-13
Remove Foxy upstream not-released file

fmauch:
since scheduled builds run on the default branch, this has to be adapted.
The file got removed on the foxy branch which is checked out by this job.

This got updated in #373


RobertWilbrandt (approved):
Makes sense

PR #398, opened 2022-06-13
Humble CI

RobertWilbrandt:
The CI setup should currently directly mirror rolling


fmauch (approved):
Thanks for adding Humble!

PR #402, opened 2022-06-14
Update README badges

fmauch:
This should be much more maintainable than this very, very overlong markdown lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,39 +33,39 @@ Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) a
   <tr>
     <th>Build Status</th>
     <td>
-      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-binary-build.yml">
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-binary-build.yml?query=event%3Aschedule++">
          <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-binary-build.yml/badge.svg?event=schedule"
               alt="Foxy Binary Build"/>
       </a> <br />
-      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-semi-binary-build.yml">
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-semi-binary-build.yml?query=event%3Aschedule++">
          <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-semi-binary-build.yml/badge.svg?event=schedule"
               alt="Foxy Semi-Binary Build"/>
       </a> <br />
-      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-source-build.yml">
-         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-source-build.yml/badge.svg?event=schedule"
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
You could consider using the non-inline version of the MD links:

```markdown
This is [my link][].

...

[my link]: http://somewhere.something/etc
```

It's certainly more readable -- perhaps even more readable than HTML.

README.md
```diff
@@ -15,11 +15,105 @@ Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) a
...
+         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-source-build.yml/badge.svg?branch=foxy"
+              alt="Foxy Source Build"/>
+      </a>
+    </td>
+    <td>
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/galactic-binary-build.yml?query=event%3Aschedule++">
+         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/galactic-binary-build.yml/badge.svg?event=schedule"
+              alt="Galactic Binary Build"/>
+      </a> <br />
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/galactic-semi-binary-build.ym?query=event%3Aschedule++l">
```
RobertWilbrandt, line 54:
```suggestion
<a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/galactic-semi-binary-build.yml?query=event%3Aschedule++">
```

RobertWilbrandt (changes requested):
This looks fine apart from the typo. I would also prefer a purely markdown-based version, but only if that would also allow splitting the links over multiple lines in the source. Therefore +1 for the HTML solution.

fmauch:
> You could consider using the non-inline version of the MD links:
> 
> ```gfm
> This is [my link][].
> 
> ...
> 
> [my link]: http://somewhere.something/etc
> ```
> 
> It's certainly more readable -- perhaps even more readable than HTML.

It doesn't get really readable once considering that those are image-embedded links. The build status line would be a very long line again:

```markdown
**Build status** | [![Foxy binary build][foxy_binary_badge]][foxy_binary_build] <br/> [![Foxy semi-binary build][foxy_semi-binary_badge]][foxy_semi-binary_build] <br/> [![Foxy source build][foxy_source_badge]][foxy_source_build] <br/> | [![Galactic binary build][galactic_binary_badge]][galactic_binary_build] <br/> [![Galactic semi-binary build][galactic_semi-binary_badge]][galactic_semi-binary_build] <br/> [![Galactic source build][galactic_source_badge]][galactic_source_build] <br/> | [![Humble binary build][humble_binary_badge]][humble_binary_build] <br/> [![Humble semi-binary build][humble_semi-binary_badge]][humble_semi-binary_build] <br/> [![Humble source build][humble_source_badge]][humble_source_build] <br/> | [![Rolling binary build][rolling_binary_badge]][rolling_binary_build] <br/> [![Rolling semi-binary build][rolling_semi-binary_badge]][rolling_semi-binary_build] <br/> [![Rolling source build][rolling_source_badge]][rolling_source_build] <br/>
```

gavanderhoorn:
Do semantic linefeeds not work for links?

I always have to put empty lines between the badges in my readme, otherwise they get rendered one after the other.

fmauch:
> Do semantic linefeeds not work for links?

They don't work inside table cells AFAIK.

Obviously, we could split the build status row into three individual rows to circumvent this... @RobertWilbrandt would you consider this a better solution?

I am somewhat undecided, as I don't consider the HTML such a bad thing...

RobertWilbrandt:
In most cases i would agree that human-readable markdown should be used, but for a table full of github links without any additional content i don't think that really applies. In this specific case i'd say that the HTML table is easier to parse for a human than equivalent markdown, and manual diffing for reviews is also easier.

RobertWilbrandt (approved):
LGTM

PR #403, opened 2022-06-14
Rework bringup

fmauch:
This PR aims at improving the startup situation for the driver. Thoughts that went into this PR:

- When installing `ur_robot_driver` it is weird that you'll have to use the `ur_bringup` package to actually start it. This could actually lead to people "rosrunning" the driver instead of using the launch files leaving them with a lot of parameters to setup.
- We currently have a dependency from the driver to the bringup package (See #366). This is only the case as you need to have the bringup package installed when you want to start the driver (also see last point).
- We've already moved the moveit startup launchfile to the `ur_moveit_config` package for similar reasons
- It does make sense to have a metapackage that you can install to have the full ecosystem around the URs installed.
- We should not directly drop the `ur_bringup` package as in galactic this kind of was the "official" way to start things. We deprecate it over the time of humble.

For now I've basically duplicated the launch and config files inside the `ur_robot_driver` package and added deprecation notices to the files in the `ur_bringup` package. The documentation is updated to use the `ur_robot_driver` package rather than the `ur_bringup` package so that new users will directly use the method that will be there for the long run. This way we can remove the package in Iron.

### DIFF
--- a/.github/workflows/ci-ros-lint.yml
+++ b/.github/workflows/ci-ros-lint.yml
@@ -18,6 +18,7 @@ jobs:
         distribution: rolling
         linter: ${{ matrix.linter }}
         package-name:
+          ur
           ur_bringup
           ur_controllers
           ur_dashboard_msgs
@@ -38,6 +39,7 @@ jobs:
           linter: copyright
           arguments: '--exclude ur_robot_driver/doc/conf.py'
           package-name:
+            ur
             ur_bringup
             ur_controllers
             ur_dashboard_msgs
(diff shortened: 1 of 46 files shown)


RobertWilbrandt (changes requested):
A few small points:
- The [README packages in the repositiory section](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/README.md#packages-in-the-repository) should contain a deprecation warning and a line for the new metapackage
- Some parts of the CI (e.g. linters) should probably include the new ```ur``` package

Otherwise, this looks really nice to me.

RobertWilbrandt (approved):
Looks good to me now

PR #404, opened 2022-06-14
Loop plan animation by default

fmauch:
In my opinion this makes a lot of sense, especially since the full path is not animated when havng this switched off.

### DIFF
--- a/ur_moveit_config/rviz/view_robot.rviz
+++ b/ur_moveit_config/rviz/view_robot.rviz
@@ -125,7 +125,7 @@ Visualization Manager:
             Show Axes: false
             Show Trail: false
             Value: true
-        Loop Animation: false
+        Loop Animation: true
         Robot Alpha: 0.5
         Robot Color: 150; 50; 150
         Show Robot Collision: false


RobertWilbrandt:
I am not sure i like this change. This diverges from the [moveit default](https://github.com/ros-planning/moveit2/blob/491f1c83da9fbae6ec4dcb16c1c24dbfaf219962/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp#L116) and the [tutorials](https://github.com/ros-planning/moveit2_tutorials/blob/80a7b61185c6b1941f0d29cf533eb3c910c209fd/doc/tutorials/quickstart_in_rviz/launch/panda_moveit_config_demo.rviz#L200) and i would prefer staying consistent to those. This avoids confusion when getting started, and when working on a custom project you probably use your own rviz config anyway.

fmauch:
> 

I always find it confusing that I only see a part of the planned motion, but if that's the default behavior also used in the tutorials I would be fine with keeping it that way...

PR #412, opened 2022-06-21
Use ros2_controllers sources

RobertWilbrandt:
ros-controls/ros2_controllers#360 already fixed a bug related to parameter handling in forward controllers. Until next sync we have to use the source repo for error-free controller startup.

### DIFF
--- a/.github/workflows/humble-execution-test.yml
+++ b/.github/workflows/humble-execution-test.yml
@@ -16,6 +16,7 @@ jobs:
         ROS_DISTRO: [humble]
         ROS_REPO: [main, testing]
     env:
+      UPSTREAM_WORKSPACE: Universal_Robots_ROS2_Driver-not-released.${{ matrix.ROS_DISTRO }}.repos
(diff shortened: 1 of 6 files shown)


fmauch (commented):
As this currently "only" affects the execution tests, we'll need some more work to make this have any effect. The execution test files don't have an upstream workspace defined.

So, in order to help the execution tests succeed, one would have to modify the workflows to use an upstream workspace and I guess, it does make sense to use the `not-released` files there...

RobertWilbrandt:
That makes sense, that would mirror the ```$ROS_DISTRO-binary-build``` configs. To be sure: This means that ```ros2_controllers``` should be in ```.$ROS_DISTRO.repos```and ```-not-released.$ROS_DISTRO.repos```, right?

fmauch:
> 

yes, for the moment that makes sense.

fmauch (approved):
LGTM!

We should remember to revert this after the next sync.

PR #414, opened 2022-06-22
Remove non-required dependency from CMakeLists

fmauch:
Similar to fd66700db47aefe7fa8f64e4068d9d5213d367d5 we need to remove that dependency on the main branch.


RobertWilbrandt (approved):
LGTM

fmauch:
Merging this, as the failing rolling testing builds seem to be unrelated.

Issue #413, opened 2022-06-21
All links on UR robot are colliding


TheFalcoGuy:
I am running the ROS2 Foxy branch of the driver and I am trying to get Moveit2 to work with my Ur5e robot. I ran the following commands in two separate terminals:

ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=192.168.0.99 launch_rviz:=false

ros2 launch ur_bringup ur_moveit.launch.py ur_type:=ur5e robot_ip:=192.168.0.99 launch_rviz:=true

Moveit launches, however, my ur5 robot is completely red: 

![Screenshot from 2022-06-21 18-37-10](https://user-images.githubusercontent.com/90654537/174908358-5ccbcd0f-800e-470d-9ae4-166b241ae0cf.png)

This results in the robot not being able to plan a trajectory. Below is my log:
[INFO] [launch]: All log files can be found below /home/d-lab/.ros/log/2022-06-21-18-36-54-686575-dlab-Alienware-m15-R6-572767
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [move_group-1]: process started with pid [572773]
[INFO] [mongo_wrapper_ros.py-2]: process started with pid [572775]
[INFO] [rviz2-3]: process started with pid [572777]
[INFO] [static_transform_publisher-4]: process started with pid [572779]
[static_transform_publisher-4] [INFO] [1655851015.113225271] [static_transform_publisher]: Spinning until killed publishing transform from 'world' to 'base_link'
[move_group-1] [WARN] [1655851015.136367207] [move_group.move_group]: MoveGroup launched without ~default_planning_pipeline specifying the namespace for the default planning pipeline configuration
[move_group-1] [WARN] [1655851015.136414473] [move_group.move_group]: Falling back to using the the move_group node namespace (deprecated behavior).
[move_group-1] Parsing robot urdf xml string.
[move_group-1] [INFO] [1655851015.138694452] [moveit_rdf_loader.rdf_loader]: Loaded robot model in 0.000740858 seconds
[move_group-1] [INFO] [1655851015.138721510] [moveit_robot_model.robot_model]: Loading robot model 'ur'...
[move_group-1] Link base had 0 children
[move_group-1] Link base_link_inertia had 1 children
[move_group-1] Link shoulder_link had 1 children
[move_group-1] Link upper_arm_link had 1 children
[move_group-1] Link forearm_link had 1 children
[...]

PR #419, opened 2022-06-23
Update build instructions to use upstream binary packages

fmauch:
The build instructions are rather outdated, as we don't need to build any upstream packages from source anymore.

The old instructions have probably lead to issues like #413


RobertWilbrandt (approved):
This seems reasonable for foxy

PR #426, opened 2022-06-28
Update linters & checkers

fmauch:
Since the main branch is targeted for ubuntu jammy, where there is no
clang-format-10 available anymore, we should raise that to 14, as this is
the default clang-format version on ubuntu jammy.


fmauch:
Since this PR became rather large I would be fine with rebase-merging this.

.pre-commit-config.yaml
```diff
@@ -65,59 +64,16 @@ repos:
...
-  # Copyright
-  - repo: local
-    hooks:
-      - id: ament_copyright
-        name: ament_copyright
-        description: Check if copyright notice is available in all files.
-        stages: [commit]
-        entry: ament_copyright
-        language: system
-
```
RobertWilbrandt, line 120: I don't really agree with removing these two for the sake of runtime reduction. I think they actually check for problems that are easily introduced by mistake and if we don't want the duplication in the ci we should tackle it there.

fmauch: That is definitively a fair point.

fmauch: I turned it the other way around in 85558b1

RobertWilbrandt (approved):
Consolidating the workflows seems like a good move. Thanks!

PR #436, opened 2022-07-13
Suggestion for update to the installation guide for URCap on a e-Series robot

vetletj:
Been trying to integrate UR10e to ROS 2 foxy with the universal robots driver, and followed [Installing a URCap on a e-Series robot](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/doc/installation/install_urcap_e_series.rst). Our UR10e did not come with Remote Control enabled out of the box, and at least from what I've seen, there isn't any guides that have pointed this out. Therefore I suggest a small update to the installation guide, where this is something the user checks before starting to install URCaps.


fmauch:
Thank you for the suggestion. from my understanding it should not necessarily be required to enable remote_control mode to move the robot, but many features will indeed require this. Maybe we can reflect this a bit more.

PR #441, opened 2022-07-19
Pin upstream ros2_controllers version to 2.9.0

fmauch:
The latest changes introduced braking changes and we require this for
building correctly. Changing from master to a specific version will make
sure that we won't suddenly not build anymore if another thing changes
upstream. This is what we have the source builds for, which indeed use
the master branch.

### DIFF
--- a/Universal_Robots_ROS2_Driver-not-released.humble.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.humble.repos
@@ -2,4 +2,4 @@ repositories:
   ros2_controllers:
     type: git
     url: https://github.com/ros-controls/ros2_controllers
-    version: master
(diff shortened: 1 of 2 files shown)


fmauch:
ToDo: Document process of handling not-released inside the files

fmauch:
@RobertWilbrandt I've added a note to the `not-released.repos` files about how I think this should be handled.

Version 2.9.0 has been synced to rolling, but not yet to humble. Should we remove it from the rolling file already in this PR or should we make a separate PR once it is synced to humble removing both entries?

fmauch:
Closing this as it is rather outdated by bow and has been superseded by #483.

PR #446, opened 2022-07-25
Add install instructions to README for released packages

michael-sbarra:
The current README has no instructions for how to install the released packages using a package manager (`apt` for Ubuntu) or with `rosdep`, so developers end up attempting to build this repo from source. This pull request attempts to mitigate that.


fmauch:
First of all thank you for the contribution!

I am not sure, whether this is the right place for that, though. In theory, the "end user" documentation would be the generated one at [docs.ros.org](https://docs.ros.org/en/ros2_packages/humble/api/ur_robot_driver/index.html). However, installation instructions are missing there, as well and I also see the "problem", that this repo will work as an entry point for many users, still.

That statement put aside, I'll add some notes to your PR.

README.md
```diff
@@ -13,6 +13,24 @@ The driver is compatible across the entire line of UR robots -- from 3 kg payloa
 Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) about this driver.
 
 
+## Installation
+
+The driver is available as several released packages. On Ubuntu, they can be installed with `apt` by running:
+
+```console
+$ sudo apt-get install ros-${ROS_DISTRO}-ur ros-${ROS_DISTRO}-ur-bringup
```
fmauch, line 21: We are currently on the way towards deprecating `ur_bringup`. Therefore, installing `ros-${ROS_DISTRO}-ur` would be sufficient. See #403 for details on that.

README.md
```diff
@@ -13,6 +13,24 @@ The driver is compatible across the entire line of UR robots -- from 3 kg payloa
 Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) about this driver.
 
 
+## Installation
```
fmauch, line 16: I think it makes sense to combine this with the compilation instructions such that you have "installation" and then as a sub chapter "Install from binary packages" and "Build from source" or similar. Also, the binary installation part should get a note, that this will not necessarily reflect the latest state on Gihub, as those binary packages are created at discrete tagged points (linked to the actual release tags page).

README.md
```diff
@@ -13,6 +13,24 @@ The driver is compatible across the entire line of UR robots -- from 3 kg payloa
...
+```
+
+The driver can also be installed with `rosdep` by added the following tags to a package manifest in a developer's workspace:
+```yaml
+# package.xml
+<depend>ur</depend>
+<depend>ur_bringup</depend>
+```
+
+Lastly, run `rosdep install --from-paths src` to complete the installation.
```
fmauch, line 31: I am not sure whether it would make sense to add this to the README. First of all, the same comment for `ur_bringup` as above is valid.

Second, depending on the actual package it might not make sense to depend on the metapackage, but rather on the package that the downstream package is actually using (`ur_robot_driver`, `ur_description`, `ur_moveit_config`...)

And third, I am afraid that this will confuse new users more than it helps them. Developers familiar with this concept will be able to infer that possibility once they know that it is available as apt package. At least that would be my assumption.

fmauch (changes requested):

fmauch:
I've just noticed, we already have this in place for the galactic branch: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/tree/galactic#install-from-binary-packages We should edit the main branch similarly.

fmauch:
Closing this in favor of #494

Issue #460, opened 2022-08-02
Driver crashing on configuring the force_torque_sensor_broadcaster


fmauch:
Currently, on the driver seems to crash on galactic when configuring the fts_broadcaster controller. See https://github.com/ros-controls/ros2_controllers/issues/404 for details

fmauch:
This has been fixed upstream.

PR #461, opened 2022-08-02
Disable fts broadcaster

fmauch:
As mentioned in #460 the current version of the driver crashes as soon as the fts_broadcaster gets started.

In order to avoid having a crashing driver during the upcoming sync period (next sync freeze will be tomorrow), I'd suggest disabling the broadcaster for now. This should leave us with a working driver except for reading FTS measurements.

I've left a note in the README explaining the behavior to users.

To minimize the review ping-pong time I've combined a patch release into this PR, as always I will push the corresponding tag to the galactic branch once this will be merged. Therefore, please merge with the rebase strategy. I'd really like to make this patch release before the next freeze.


fmauch:
This might not be necessary if we manage to get https://github.com/ros-controls/ros2_control/pull/795 in time.

fmauch:
Closing as this is unnecessary as mentioned above.

Issue #390, opened 2022-06-09
scaled JTC does not work on fake_hw


fmauch:
When trying to use the scaled JTC on Fake HW I get 

```
[ros2_control_node-1] terminate called after throwing an instance of 'std::underflow_error'
[ros2_control_node-1]   what():  addition leads to int64_t underflow
```

Obviously this needs to be further debugged, but I wanted to post it here as a first step.

fmauch:
Update: It seems that all floating point io states generate a `nan` value when using fake_hw. @destogl do you know where this could come from? For example, the output of the `/io_and_status_controller/tool_data` is

```
...
tool_output_voltage: 0
tool_current: .nan
tool_temperature: .nan
tool_mode: 0
```

ferrolho:
I think we can remove the "galactic:" prefix from the title ─ I just tested this on ROS 2 Humble, and I am getting the same error message:
```
[ros2_control_node-1] terminate called after throwing an instance of 'std::underflow_error'
[ros2_control_node-1]   what():  addition leads to int64_t underflow
```

fmauch:
@ferrolho thanks for adding this information

haider8645:
This seems to be still open. I had to switch the SJTC to JTC on humble.

fmauch:
the scaled jtc should work on mock hardware (fake hardware on humble) by now.

gautz:
@fmauch 
Trying to [use the mock hardware on Rolling](https://docs.ros.org/en/ros2_packages/rolling/api/ur_robot_driver/usage.html#start-hardware-simulator-or-mockup)
`ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_mock_hardware:=true launch_rviz:=true`
I'm still getting a similar error:
```
[INFO] [launch]: All log files can be found below /home/etudiant/.ros/log/2024-06-07-11-51-40-145410-ubuntu22ros2-3870
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [ros2_control_node-1]: process started with pid [3876]
[INFO] [robot_state_publisher-2]: process started with pid [3878]
[INFO] [rviz2-3]: process started with pid [3880]
[INFO] [spawner-4]: process started with pid [3882]
[INFO] [spawner-5]: process started with pid [3884]
[INFO] [spawner-6]: process started with pid [3886]
[...]

Issue #301, opened 2022-02-23
Galactic: Controller example Can't activate controller 'scaled_joint_trajectory_controller'


lmendyk:
When following the example I get the following error, which is the reason for  move it example to fail execute a plan.

`ros2 launch ur_bringup ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=false
[INFO] [launch]: All log files can be found below /home/lmendyk/.ros/log/2022-02-23-11-14-26-240247-lmendykLinux-17283
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [ros2_control_node-1]: process started with pid [17287]
[INFO] [robot_state_publisher-2]: process started with pid [17289]
[INFO] [spawner-3]: process started with pid [17291]
[INFO] [spawner-4]: process started with pid [17293]
[INFO] [spawner-5]: process started with pid [17297]
[INFO] [spawner-6]: process started with pid [17306]
[INFO] [spawner.py-7]: process started with pid [17315]
[INFO] [spawner-8]: process started with pid [17321]
[ros2_control_node-1] [INFO] [1645611267.502405838] [controller_manager]: update rate is 500 Hz
[robot_state_publisher-2] Link base_link had 2 children
[robot_state_publisher-2] Link base had 0 children
[robot_state_publisher-2] Link base_link_inertia had 1 children
[robot_state_publisher-2] Link shoulder_link had 1 children
[robot_state_publisher-2] Link upper_arm_link had 1 children
[robot_state_publisher-2] Link forearm_link had 1 children
[robot_state_publisher-2] Link wrist_1_link had 1 children
[robot_state_publisher-2] Link wrist_2_link had 1 children
[robot_state_publisher-2] Link wrist_3_link had 2 children
[robot_state_publisher-2] Link flange had 1 children
[robot_state_publisher-2] Link tool0 had 0 children
[robot_state_publisher-2] Link ft_frame had 0 children
[robot_state_publisher-2] [INFO] [1645611267.485396000] [robot_state_publisher]: got segment base
[robot_state_publisher-2] [INFO] [1645611267.485565494] [robot_state_publisher]: got segment base_link
[robot_state_publisher-2] [INFO] [1645611267.485588164] [robot_state_publisher]: got segment base_link_inertia
[...]

PR #464, opened 2022-08-11
add ur_moveit.launch.py parameter to use working controller when using fake hardware

adverley:
Workaround for #390 (simulated hardware throws an underflow error which breaks scaled_joint_trajectory_controller #301) by allowing user to specify whether the robot driver is using fake_hardware as CLI argument. If this is the case, MoveIt will fall back on the joint_trajectory_controller. This behavior is disabled by default and requires user to manually set argument is_using_fake_hardware to true when running ur_moveit.launch.py.

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -45,6 +45,7 @@ def launch_setup(context, *args, **kwargs):
 
     # Initialize Arguments
     ur_type = LaunchConfiguration("ur_type")
+    is_using_fake_hardware = LaunchConfiguration("is_using_fake_hardware")
     safety_limits = LaunchConfiguration("safety_limits")
     safety_pos_margin = LaunchConfiguration("safety_pos_margin")
     safety_k_position = LaunchConfiguration("safety_k_position")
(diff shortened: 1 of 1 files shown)


fmauch:
Thank you @adverley for the contribution!

I like the idea of that approach, but I think it would be nice to keep the argument name consistent, which would be `use_fake_hardware`

adverley:
Good for me, I changed the arguments name back to `use_fake_hardware`. I also let pre-commit format the file. Should be good now.

ur_moveit_config/launch/ur_moveit.launch.py
```diff
@@ -162,6 +163,11 @@ def launch_setup(context, *args, **kwargs):
 
     # Trajectory Execution Configuration
     controllers_yaml = load_yaml("ur_moveit_config", "config/controllers.yaml")
+    # the scaled_joint_trajectory_controller does not work on simulated robots
```
fmauch, line 166:
```suggestion
    # the scaled_joint_trajectory_controller does not work on fake hardware
```

This does not necessarily affect other simulations such as URsim or gazebo.

fmauch:
I've just tested this and it seems to work fine.

ferrolho:
Hi, all! Currently, on the _main_ branch (Humble), the _scaled_joint_trajectory_controller_ is not working with `use_fake_hardware:=true`, right? Apart from testing it on the real robot , is there any other way to test that controller?

fmauch (approved):
LGTM the way this is now. Thank you @adverley

fmauch:
> Apart from testing it on the real robot , is there any other way to test that controller?

You can use URSim to simulate a real robot. from a driver's perspective this will not make any difference than a real robot. See [the docs](https://docs.ros.org/en/ros2_packages/humble/api/ur_robot_driver/installation/ursim_docker.html) for details on that.

Issue #420, opened 2022-06-27
Reverse IP parameter missing for ROS2 driver


fmauch:
In certain situations (e.g. driver running in a docker container) it is useful to specify the IP address injected into the script code where the reverse interface is reachable from the robot. (See for example https://forum.universal-robots.com/t/real-robot-is-not-moving-with-moveit/22183/3)

The ROS2 version currently does not support setting this manually.

fmauch:
This has been implemented in #466

Issue #447, opened 2022-07-26
Zeroing the F/T Sensor?


danzimmerman:
Is there a built-in way to zero the F/T sensor from the driver to get rid of no-load offsets before carrying out operations that use the sensor? I get a few Newtons of force with no tool or payload, within the +/- 4N spec but I'd like to zero it out from time to time.

I know there's a `zero_ft_sensor()` URScript command but
* Looking at the features docs [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/doc/features.rst) it appears we can't send isolated script commands.
* Looking at `/ur_robot_driver/resources/ros_control.urscript` (referenced in launch [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/launch/ur_control.launch.py#L84)) it seems like the running program is just handling the speed or position commands.
* It doesn't seem like `/dashboard_client` or `/io_and_status_controller` provide anything to do this.

In the absence of a way to call `zero_ft_sensor()` I can subscribe to `/force_torque_sensor_broadcaster/wrench` for a while and subtract an average offset wrench from the measured wrench. Maybe I could also zero the in my Polyscope robot program prior to the External Control node running?

However, I wanted to check that I'm not missing something in the available interfaces.

fmauch:
This is correct. I'll add this issue to the ROS1 feature parity Milestone. This is directly depending on #251.

PR #316, opened 2022-03-21
Default path to ur_client_library urscript

livanov93:
PR changes default urscript that is sent to the robot when `headless_mode` is used.

PR #466, opened 2022-08-18
Added services to set tool voltage and zero force torque sensor

urmahp:
Added launch arguments for reverse ip and script command interface port.

This requires the changes from [pr](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/111) in the client library and also the changes from [pr](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/38) in the ur ros2 description. 

To be able to use the script commands when the robot is in local control we have to use the urscript from the client, this is being addressed in #316.  

This will fix the issues #420 and #447.

### DIFF
--- a/ur_controllers/include/ur_controllers/gpio_controller.hpp
+++ b/ur_controllers/include/ur_controllers/gpio_controller.hpp
@@ -63,16 +63,19 @@ enum CommandInterfaces
 {
   DIGITAL_OUTPUTS_CMD = 0u,
   ANALOG_OUTPUTS_CMD = 18,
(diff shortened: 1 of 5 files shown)


fmauch:
I've added the script from the library, as the updates are needed and we don't have to wait for #316.

dpiet:
Is there something preventing this PR from being merged?

fmauch:
> Is there something preventing this PR from being merged?

For instance the failing format pipeline. And time to fix this / do some rework since #316 is merged by now.

fmauch (approved):
LGTM and tested. Thanks @urmahp

Issue #434, opened 2022-07-13
Issue to build Driver on Humble


LukGen:
Hello,

I have trouble to build the ur-driver. Here is my console output:

colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
[0.230s] WARNING:colcon.colcon_core.package_selection:Some selected packages are already built in one or more underlay workspaces:
	'gripper_controllers' is in: /opt/ros/humble
	'joint_trajectory_controller' is in: /opt/ros/humble
	'joint_state_broadcaster' is in: /opt/ros/humble
If a package in a merged underlay workspace is overridden and it installs headers, then all packages in the overlay must sort their include directories by workspace order. Failure to do so may result in build failures or undefined behavior at run time.
If the overridden package is used by another package in any underlay, then the overriding package in the overlay must be API and ABI compatible or undefined behavior at run time may occur.

If you understand the risks and want to override a package anyways, add the following to the command line:
	--allow-overriding gripper_controllers joint_state_broadcaster joint_trajectory_controller

This may be promoted to an error in a future release of colcon-override-check.
Starting >>> forward_command_controller
Starting >>> joint_trajectory_controller
Starting >>> force_torque_sensor_broadcaster
Starting >>> joint_state_broadcaster
Starting >>> ur_dashboard_msgs
Starting >>> ur_description
Starting >>> ur_msgs
Starting >>> ros2_controllers_test_nodes
Finished <<< ros2_controllers_test_nodes [2.03s]                                
Starting >>> ur_client_library
Finished <<< ur_description [2.77s]                                             
Starting >>> diff_drive_controller
--- stderr: joint_state_broadcaster                                                   
In file included from /home/developer/ros_ur_driver/src/ros2_controllers/joint_state_broadcaster/src/joint_state_broadcaster.cpp:15:
/home/developer/ros_ur_driver/src/ros2_controllers/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp:74:25: error: ‘CallbackReturn’ in namespace ‘controller_interface’ does not name a type
   74 |   controller_interface::CallbackReturn on_init() override;
      |                         ^~~~~~~~~~~~~~
/home/developer/ros_ur_driver/src/ros2_controllers/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp:77:25: error: ‘CallbackReturn’ in namespace ‘controller_interface’ does not name a type
[...]

PR #470, opened 2022-09-02
back-port use_fake_hardware parameter in ur_moveit.launch.py to Galactic (#464)

haavarpb:
The patch in #464 fixes my problem where the UR Gazebo Sim won't move because the Scaled JTC controller is loaded into the MoveitSimpleControllerManager as default controller (and there is no way to change it during runtime through e.g. a service). Building main branch breaks because of same error message as in #434, so I found it simpler to backport a single commit from main.

This is my very first pull request in a professional repository so sorry in advance for any mistakes or shortages. For example, I am not sure which tests you refer to under item 3 in the [contributing page](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/f580ae73edc67b10ce89ff72d30f70269935e39d/CONTRIBUTING.md): Are you referring to the pre-commit tests?

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -43,6 +43,7 @@ def launch_setup(context, *args, **kwargs):
 
     # Initialize Arguments
     ur_type = LaunchConfiguration("ur_type")
+    use_fake_hardware = LaunchConfiguration("use_fake_hardware")
     safety_limits = LaunchConfiguration("safety_limits")
     safety_pos_margin = LaunchConfiguration("safety_pos_margin")
     safety_k_position = LaunchConfiguration("safety_k_position")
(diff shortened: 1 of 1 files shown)


destogl (approved):
The PR seems fine.

The referenced backport PR seems to not be correct.

grassjelly:
Isn't `use_fake_hardware` a non boolean type which will always return True when evaluated with `use_fake_hardware`?

destogl:
> Isn't `use_fake_hardware` a non boolean type which will always return True when evaluated with `use_fake_hardware`?

That is true, I am not sure this can be used as proposed because of the ROS2 launch system.

@haavarpb have you tested this?

fmauch:
> Isn't use_fake_hardware a non boolean type which will always return True when evaluated with use_fake_hardware?

Yes, absolutely, I was currently running into this also on the main branch. Looks like I did a poor job when testing #464. Probably I only checked the case setting fake_hw to `true` where the result is as expected...

Therefore, this should be fixed first, but then we can definitively port this feature to the galactic branch, as well.

grassjelly:
this worked for me

fmauch:
@grassjelly what is the `servo_yaml["command_out_topic"] = "/joint_trajectory_controller/joint_trajectory"` for?

grassjelly:
It's to switch the realtime servoing's output command when use_fake_hardware is set to `True` https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_moveit_config/config/ur_servo.yaml#L62 and connect with the default controller on fake hardware.

fmauch:
I see. That's definitively something worth debating, but I'd leave that out of this context, then, as it by default uses the forwarding controller, which should work on both fake, and non-fake hw (given the controller is activated)

fmauch:
This should be adapted according to #480

fmauch:
@haavarpb please cherry=pick 6c52828f6dcb4e736952d0fae5fd99522835ab4a to this PR in order to finish this.

fmauch (approved):
LGTM now

Issue #252, opened 2021-12-02
Invalid path in git


michaelaeriksen:
When checking out this repo I get 

OS: Windows 

```
error: invalid path 'ur_robot_driver/resources/ursim_driver/ursim/aux/.vol/urcap_ros_control.installation'
```

Note: The ROS1 version of these driver do not have this issue.

urrsk:
@michaelaeriksen Thanks for your question.
I am not really sure why windows is not happy with that filename, but thanks for the insight
Maybe you can get going with Windows Subsystem for Linux (WSL). 

Please note that the driver does not yet support Windows natively.

michaelaeriksen:
I forked the `Universal_Robots_ROS2_Driver` repo in wsl and removed the files giving me problems in Windows. I have the ROS version of the drivers working on Windows natively and I'm now looking at migrating to ROS2. A while back I forked the `Universal_Robots_Client_Library` repo and changed all the socket code to use the platform independent library asio and the ROS version of the drivers are working great.

urrsk:
@michaelaeriksen that sounds great. Have you any idea of why windows did not like that file?
We have talked about making the ROS2 driver windows compatible, and here you work sounds very interesting.
Is it possible for you to share the ```asio``` version?

relffok:
`aux` seems to be a reserved filename on windows. We ran into this issue before and renaming the path worked for us! 

https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

fmauch:
> `aux` seems to be a reserved filename on windows. We ran into this issue before and renaming the path worked for us!
> 
> https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

Thanks for that insight!

haellingsen:
How did you workaround this? Did you fork and rename?

fmauch:
@haellingsen you could use the branch from #478 for example. This will see some more work today, but the problematic folder is already removed.

appselskapet:
> @haellingsen you could use the branch from #478 for example. This will see some more work today, but the problematic folder is already removed.

Thanks.

Although I ran git config core.protectNTFS false and cloned it.

fmauch:
Since the path is no longer part of the main branch, I'm going to close this.

PR #478, opened 2022-09-21
Remove the custom ursim docker files

fmauch:
This has been migrated inside the docs and is not needed anymore.

Solves https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/252


urmahp (approved):
I think it is a good idea to remove the ursim_driver and make the user use the official docker image instead. It is also fine with me to remove the Dockerfile, since it was outdated anyways.

PR #482, opened 2022-09-23
Fixed argument/parameter usage of urdf macro

p0rys:
"use_fake_hardware" and "fake_sensor_commands" are used as arguments instead of params. This required external urdf files that will use these macro to declare these two values as arguments instead of passing them via marcro parameter.

### DIFF
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -85,9 +85,9 @@
     <!-- ros2 control instance -->
     <xacro:ur_ros2_control
       name="URPositionHardwareInterface" prefix="${prefix}"
-      use_fake_hardware="$(arg use_fake_hardware)"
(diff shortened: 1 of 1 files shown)


fmauch:
This is not only affecting those. This has been fixed from galactic up, but was not backported to Foxy, yet. Thanks for pointing that out! The Foxy description seems to need some more attention....

p0rys:
Ah i see, i was only working with the Foxy branch because its our main ROS branch right now and i had a tight deadline (as always ;D) 

Are there any galatic / humble specific changes within the new description repository or is it maybe possible to just use the new package as a submodule in Foxy too?

fmauch:
As far as I remember, there were some major changes, especially regarding the ros2_control integration, so I wouldn't expect that you can simply use that on a Foxy setup.

fmauch:
Closing in favor of #495

PR #483, opened 2022-09-23
Remove pinning ros2_control to specific branch

### LINKED ISSUE #442: Revert ros2_control upstream version to released one once 2.9.0 is released
We currently need 2.9.0 to build against, but this is not yet synced to the main repos. Once this is synced, we can remove the ros2_controllers entry from the not_released.repos files.

See #442

fmauch:
This makes #441 obsolete and fixes #442


urmahp (approved):
I have approved the PR, since it is only a very small typo adjustment. I have verified that it is possible to build and launch driver with the changes.

PR #490, opened 2022-10-04
Fixes launchfile references

fmauch:
The filename with the dual type ending was wrong.

Thanks to @urmahp for finding this inside #451


urmahp (approved):
LGTM

PR #492, opened 2022-10-05
Adapt jtc controller params to new param api

fmauch:
This should fix our pipelines for the current testing repo.


fmauch:
Converting to draft as the breaking changes have not been released, yet. Once released, we can add the ros2_controllers version to our `not-released` files and then prepare a release ourselves.

fmauch:
I've disabled caching, as this seemed to cause problems. As `ros2_controllers` had not been built from source before, the cached workspace was referring to ros2_controllers from the installed packages. When starting with an empty workspace, this uses the upstream version correctly.

I suggest to keep all three individual commits either by rebasing or commit merging, as thy are not necessarily overlapping and since da50cf4 will get reverted after the next sync.

RobertWilbrandt (approved):
Unfortunate that we have to disable caching, but seems like this is the best option for now

PR #495, opened 2022-10-05
Move all arguments out of macro file

fmauch:
This way it can be included more easily.

@p0rys this should fix the problems mentioned in #482.

### DIFF
--- a/ur_description/urdf/ur.urdf.xacro
+++ b/ur_description/urdf/ur.urdf.xacro
@@ -23,6 +23,11 @@
    <xacro:arg name="initial_positions_file" default="$(find ur_description)/config/initial_positions.yaml"/>
    <xacro:property name="initial_positions_file" default="$(arg initial_positions_file)"/>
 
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Would be even better if there would be a way to keep the `ros2_control` block out of the default macro completely.

It's really rather awkward having that present in all cases. The changes in this PR seem to also highlight that.

fmauch:
> Would be even better if there would be a way to keep the `ros2_control` block out of the default macro completely.
> 
> It's really rather awkward having that present in all cases. The changes in this PR seem to also highlight that.

I do agree on the awkward part, but so far I couldn't come up with a satisfying solution. I've started #496 to share my concerns. Maybe we can discuss things there a bit to migrate to a proper solution?

fmauch:
@RobertWilbrandt a review on this would be appreciated before we put Foxy to its EOL sleep.

fmauch:
Closing this, since Foxy is no longer updated.

PR #496, opened 2022-10-05
Move out ros2_control macro from ur kinematic macro

fmauch:
This PR aims at taking a step further in refactoring the description, as its current state is rather suboptimal. With this PR I strive for
 - Having a separated visual description from the control description
 - Having a single macro that users can include

I'll start this PR as a draft as I'd like to incorporate feedback from the community on how to structure things. In the long run I'd like to apply this restructuring also to the Humble and Rolling versions.

### DIFF
--- a/ur_description/urdf/ur.urdf.xacro
+++ b/ur_description/urdf/ur.urdf.xacro
@@ -4,7 +4,6 @@
    <xacro:arg name="name" default="ur"/>
    <!-- import main macro -->
    <xacro:include filename="$(find ur_description)/urdf/ur_macro.xacro"/>
-   <xacro:include filename="$(find ur_description)/urdf/inc/ur_common.xacro" />
 
    <!-- parameters -->
    <xacro:arg name="prefix" default="" />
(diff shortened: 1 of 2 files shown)


fmauch:
With that approach we've separated the robot's kinematics from the control xacro part while still providing a single macro to include. I'm not completely happy with this for the following reasons:
- When creating their own descriptions users would either include the `ur_macro.xacro` part and instanciate `xacro:ur_macro` or include  `ur_kinematics_macro.xacro` and instanciate `xacro:ur_kinematics`. This would however require to also parse the config files beforehand using the `xacro:read_model_data` macro. While that offers more flexibility, this also increases complexity by adding another macro layer.
- The parsed config file data is used in both, the kinematics description and the control description. I am not sure whether we could call `xacro:read_model_data` inside both macros, that might be an easy way to mitigate this.
- I don't really see the benefits, but that's probably either because I'm thinking about this the wrong way or that I don't have certain use cases in mind.

fmauch:
I'm closing this, as this will not be implemented for foxy anymore. The discussion about this still goes on in https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/56

PR #512, opened 2022-10-11
Clean up & improve execution tests

RobertWilbrandt:
The goal of this PR are:
- [X] Clean up the remnants of ROS1 tests stil present in the driver
- [X] Start ursim as part of the execution test
- [x] Get rid of custom docker images in favor of upstream ursim
- [x] Clean up tests so they are reasonably maintainable


ur_robot_driver/test/dashboard_client.py
```diff
@@ -0,0 +1,202 @@
...
+        cls.node = Node("dashboard_client_test")
+        cls.init_robot(cls)
+
+    @classmethod
+    def tearDownClass(cls):
+        # Shutdown the ROS context
+        cls.node.destroy_node()
+        rclpy.shutdown()
+
+    def init_robot(self):
```
fmauch, line 121: Intuitively I would say that this should be a `classmethod`, as well.

fmauch (commented):
Only minor comments from my side except for the testing structure. I think we should drop the separate execution tests.

fmauch:
Oh, and the `github/dockerursim` folder could probably be deleted in this PR, as well.

.github/workflows/humble-execution-test.yml
```diff
@@ -17,17 +17,13 @@ jobs:
         ROS_REPO: [main, testing]
     env:
       UPSTREAM_WORKSPACE: Universal_Robots_ROS2_Driver-not-released.${{ matrix.ROS_DISTRO }}.repos
-      DOCKER_RUN_OPTS: --network static_test_net
+      DOCKER_RUN_OPTS: --network ursim_net
```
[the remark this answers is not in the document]
RobertWilbrandt: You're right, that was an oversight on my part

.github/workflows/humble-execution-test.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   binary_test:
-    name: binary execution test
+    name: ${{ matrix.ROS_DISTRO }} (${{ matrix.ROS_REPO }}) binary execution test
```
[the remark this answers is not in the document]
RobertWilbrandt: According to the [industrial_ci example](https://github.com/ros-industrial/industrial_ci/blob/master/doc/industrial_ci_action.yml#L17) this should work

.github/workflows/humble-execution-test.yml
```diff
@@ -7,7 +7,7 @@ on:
```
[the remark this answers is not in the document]
RobertWilbrandt: I agree, i'll adjust the PR for this

ur_robot_driver/test/dashboard_client.py
RobertWilbrandt: i don't think i agree. A class instance holds the node and all the service/action interfaces for the driver, so the initialisation should also be bound to that instance. It is unfortunate that some detail of the class initialization before `setUpClass` leads to the weird `self.call_service(self, ...` syntax, but i don't see how this would motivate lifting this to a `classmethod`.

RobertWilbrandt:
`dockerursim` is still used by galactic, so it cannot be removed now. As galactic EOL is already pretty close i would not touch that for now but remove it afterwards.

ur_robot_driver/test/dashboard_client.py
fmauch: Since it is called from a class method I still thing this is weird, but I can live with that for the moment.

fmauch (approved):
LGTM, Thank you @RobertWilbrandt!

RobertWilbrandt:
The humble binary failure should be fixed with the next humble sync. I think this is ready to get merged in.

.github/workflows/humble-source-build.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/Universal_Robots_ROS2_Driver.${{ env.ROS_DISTRO }}.repos
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
-          skip-tests: true
+          extra-cmake-args: -DUR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=ON
```
fmauch, line 40:
```suggestion
```

I would not expect the integration tests to run here, since we are again building inside a docker container. You can mount the docker socket into this container, but [according to the docs](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container#setting-container-resource-options) it is not possible to set the network of that container, as you did with the industrial_ci.

In my opinion that's fine to omit the integration tests here, as the source builds aim for more long-term API incompatibilities.

.github/workflows/rolling-source-build.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/Universal_Robots_ROS2_Driver.${{ env.ROS_DISTRO }}.repos
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
-          skip-tests: true
+          extra-cmake-args: -DUR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=ON
```
fmauch, line 40:
```suggestion
```
Same reasoning as for humble_source build

ur_robot_driver/CMakeLists.txt
```diff
@@ -156,6 +162,15 @@ if(BUILD_TESTING)
   find_package(ur_description REQUIRED)
   find_package(ur_msgs REQUIRED)
   find_package(launch_testing_ament_cmake)
-  add_launch_test(test/integration_test_1.py)
-  add_launch_test(test/integration_test_2.py)
+
+  if(${UR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS})
```
fmauch, line 166: A comment for clarity is probably a good thing to do here.
```suggestion
  # The integration tests will most probably not run inside the buildfarm, since we need to be able to create new docker 
  # containers and provide the ursim_net for the container running the CI which we cannot influence
  if(${UR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS})
```

fmauch (changes requested):
I think we should adapt the source workflows as explained in the comments, as I am pretty sure they will not work this way.

.github/workflows/humble-source-build.yml
RobertWilbrandt: I am not quite convinced that this doesn't work. The problem in the (semi-)binary builds was that we had a setup like this:
```
Github container
  |-> ros_industrial container
    |-> ursim
```
In my understanding, mounting the socket and setting the network turned that into this:

```
Github container
  |-> ros_industrial container (network=ursim_net)
  |-> ursim (network=ursim_net)
```
This worked fine. In the scenario here, however, we only have the following:
```
Github container
  |-> ursim (network=ursim_net)
```
with the tests running directly in the github runner. I don't see how this would require any additional configuration to work, and if there is no technical reason to not have these tests run i would prefer to have them for all test configurations.

fmauch: I am not sure whether the deduced structure is correct, as by defining the container an additional docker run step is added in the workflow run. You might as well also be correct though.

RobertWilbrandt: I temporarily added a commit that runs the source tests on pull requests to check this behavior

fmauch:
Our tests seem still to be randomly timing out... Which is odd, since the tests should only take about 1.5 minutes, as seen [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/runs/3536819460/jobs/5936232806#step:4:2833) and our timeout is currently at 500 seconds.

fmauch (commented):
Overall this seems to be in a fine and mergable state. @RobertWilbrandt do you think we should squash all commits or would you like to clean up the history a bit?

PR #514, opened 2022-10-12
publisher_joint_trajectory_position_controller nodes name changed on launch and yaml files in the bringup package it was not working before

mertgungor:


fmauch (approved):
Thanks @mertgungor for pointing that out!

PR #524, opened 2022-10-23
Add a note in the tool_comm doc about a URCap conflict

### LINKED ISSUE #523: Add warning about tool_comm conflict with Robotiq URCap
As discussed in https://github.com/UniversalRobots/Universal_Robots_ToolComm_Forwarder_URCap/issues/9 there seems to be a conflict between the tool forwarding URCap and e.g. the Robotiq-Gripper URCap. As suggested there, we should warn users about this in [our documentation](https://docs.ros.org/en/ros2_packages/rolling/api/ur_robot_driver/setup_tool_communication.html).

fmauch:
This should resolve #523

### DIFF
--- a/ur_robot_driver/doc/setup_tool_communication.rst
+++ b/ur_robot_driver/doc/setup_tool_communication.rst
@@ -4,9 +4,9 @@ Setting up the tool communication on an e-Series robot
 ======================================================
 
 .. note::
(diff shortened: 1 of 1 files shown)


urmahp (approved):
LGTM, but I have added a small suggestion.

PR #526, opened 2022-10-26
Fix tool voltage setup

### LINKED ISSUE #521: Should the default tool voltage be zero?
I'm digging into tool communication and realize that I plugged a gripper into the robot flange and ran the driver without really registering the idea that it could change the tool voltage from software, and that this voltage defaults to 24V:

https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/launch/ur_control.launch.py#L533

This also applies to the robot description:

https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/ros2/urdf/ur.ros2_control.xacro#L16

I'm not even sure the `tool_voltage` parameter is active yet, maybe that's coming in with #466, and the **actual** tool voltage is still set to whatever it is inside Polyscope?

image

I suppose 12V tools are rare for power delivery reasons, and maybe have overvoltage protection anyway, but defaulting to zero and documenting that it needs to be configured might prevent some accidents for users with 12V tools.

fmauch:
Fixes #521

This basically adds https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/101

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -530,7 +530,7 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument(
             "tool_voltage",
-            default_value="24",
+            default_value="0", # 0 being a conservative value that won't destroy anything
             description="Tool voltage that will be setup.",
         )
     )
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -530,7 +530,7 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument(
             "tool_voltage",
-            default_value="0", # 0 being a conservative value that won't destroy anything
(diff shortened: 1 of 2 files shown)


ur_robot_driver/launch/ur_control.launch.py
```diff
@@ -530,7 +530,7 @@ def generate_launch_description():
     declared_arguments.append(
         DeclareLaunchArgument(
             "tool_voltage",
-            default_value="24",
+            default_value="0", # 0 being a conservative value that won't destroy anything
```
urmahp, line 533: I think that a space is missing before the comment to make the formatter happy.

urmahp (approved):
LGTM other than small comment. 

Is it the plan to use the control script from the client library as we do in the ROS1 driver?

ur_robot_driver/launch/ur_control.launch.py
fmauch: > Is it the plan to use the control script from the client library as we do in the ROS1 driver?

See #316

PR #544, opened 2022-11-22
Adapt ci for humble branch

fmauch:
Since we branched out for Humble, there are a couple of things to adapt on the main branch. This PR aims at covering all of them.


.github/workflows/humble-binary-build.yml
```diff
@@ -1,11 +1,5 @@
 name: Humble Binary Build
 on:
-  pull_request:
-    branches:
-      - main
-  push:
-    branches:
-      - main
```
RobertWilbrandt, line 8: Why do we no longer want to test on these? This is inconsistent with foxy and galactic which don't have this disabled

RobertWilbrandt: Oh never mind, i looked at the wrong branch. This is correct, as we only run the scheduled cis on main (apart from rolling) and the `pull_request`/`push` ones on their respective branches

RobertWilbrandt:
Thanks, this seems consistent with our handling of the other branches

RobertWilbrandt (approved):

PR #546, opened 2022-11-23
Link stdc++fs

fmauch:
This seems to be necessary for older compilers (e.g. gcc 8.5 on RHEL8)

See e.g. https://build.ros2.org/job/Hbin_rhel_el864__ur_calibration__rhel_8_x86_64__binary/23/console

This might be necessary for rolling, as well, but for rolling a lot of upstream dependencies are missing on RHEL before this gets relevant.

It might be beneficial to add RHEL builds to our pipeline based on https://github.com/ros-controls/ros2_rhel


RobertWilbrandt:
This contains a lot of changes that are not related at all to what you are describing. Did you mangle this with some other PR?

fmauch:
I've been playing around with an RHEL workflow, but it seems to be failing on installing dependencies that should be there, see https://github.com/fmauch/Universal_Robots_ROS2_Driver/actions/runs/3536558367/jobs/5935691095.

PR #551, opened 2022-11-30
Document cloning the correct branch for humble

RobertWilbrandt:
This seems to have been forgotten when branching out humble.


fmauch (approved):
thanks for catching that!

PR #552, opened 2022-11-30
Clean up & improve execution tests (Backport of #512)

fmauch:
This is basically the backport of #512 

This should make PRs onto this branch run the tests as well and make the scheduled builds use the correct tests for humble, as well.


RobertWilbrandt (approved):
This seems complete to me. Thank you!

PR #567, opened 2022-12-09
Support for multi arm setups

firesurfer:
This PR is intended to include multiple commits that will enable support for multiple arms in the end.

See https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/47

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -273,24 +273,31 @@ URPositionHardwareInterface::on_activate(const rclcpp_lifecycle::State& previous
   RCLCPP_INFO(rclcpp::get_logger("URPositionHardwareInterface"), "Starting ...please wait...");
 
   // The robot's IP address.
-  std::string robot_ip = info_.hardware_parameters["robot_ip"];
+  const std::string robot_ip = info_.hardware_parameters["robot_ip"];
   // Path to the urscript code that will be sent to the robot
-  std::string script_filename = info_.hardware_parameters["script_filename"];
+  const std::string script_filename = info_.hardware_parameters["script_filename"];
   // Path to the file containing the recipe used for requesting RTDE outputs.
-  std::string output_recipe_filename = info_.hardware_parameters["output_recipe_filename"];
+  const std::string output_recipe_filename = info_.hardware_parameters["output_recipe_filename"];
   // Path to the file containing the recipe used for requesting RTDE inputs.
-  std::string input_recipe_filename = info_.hardware_parameters["input_recipe_filename"];
+  const std::string input_recipe_filename = info_.hardware_parameters["input_recipe_filename"];
   // Start robot in headless mode. This does not require the 'External Control' URCap to be running
   // on the robot, but this will send the URScript to the robot directly. On e-Series robots this
   // requires the robot to run in 'remote-control' mode.
-  bool headless_mode =
+  const bool headless_mode =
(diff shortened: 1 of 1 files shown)


fmauch:
First of all thank you for all the effort you put into this and for contributing this back. From a brief look, this seems to be rather good! Unfortunately my time will be rather restricted this week and I will need some lab time to test this so it will probably have to wait until next week unless @RobertWilbrandt wants to have a peek at this before :-)

firesurfer:
So with the latest commits I managed to get it running in a setup with two UR16e robot arms.  
The major trick there (apart from using non blocking reads) was to start up the rtde communcation at the latest point possible. (Actually when the first read is done)

In my test I used the standard ros2control node btw.

firesurfer:
@fmauch what is the current state? If you need any help getting this to work on your setup let me know

firesurfer:
Hi hope you are doing fine again!

For our workflow here it would be really beneficial to get this merged at some point. So if I can do anything to help getting this merged let me know ;)

Regarding #568 - For stable operation with having two robot arms up and running and using the standard ros2controller node I had to take care of the `joint_state_broadcaster` and also hack in the rt priority setting in the ur driver (see my latest commit). The issue with the standard `joint_state_broadcaster` is that it will publish to `/joint_states` with the update rate of the controllermanager (which is 500Hz for the arms). Publishing joint states at 500Hz rate will result into a huge load at other point of the system (such as the robot_state_publisher and moveit, and for some reason also rviz2). I solved this by having the `joint_state_broadcaster` publish to a different topic and have a node in between that republish to `/joint_states`  at a lower rate (100Hz is for our use case more than enough). 

Reducing the overall systemload makes it much easier to avoid dropping the connection because of a too low update rate.

firesurfer:
Quick update. I also added a field to set the allowed keep alive counter. At least for our setup it is currently not necessary to reach exactly the 500Hz update rate.

fmauch:
I think it would be good to break this PR into multiple smaller ones. Many of them are not necessarily related towards multi-arm setup. Some of the changes have also been merged already, e.g. in #466.

By quickly skimming over this, I suggest splitting this into at least
 - One PR adding all missing port configurations (if there are any left)
 - One PR migrating to generate_parameter_library
 - One PR adding the prefix (Is that needed at this level? I haven't looked into that too deep)

If you want @firesurfer we can create a milestone to tie all of them together.

firesurfer:
Hi @fmauch  sorry for my late answer. I was pretty sick the last  few days unfortunately.

I will try to split up this PR into multiple single ones in the next days.

firesurfer:
@fmauch I split up the PR into 4  parts. 

The generate_parameter_libary and the prefix are still in one PR as seperating them is quite a bit of work (see the related PR) 

Regarding the prefix:

I decided to go for the prefix as this is the less work intensive way of achieving unique interface names. In the joint_trajectory_controller this is solved for example by passing the full joint_names (including their prefix from the urdf) to the controller. As the gpio controller is very specific to the ur arm using the prefix should be fine. 

The main point one might want to argue is whether the prefix should be used for the resulting controller topics. 
I did this because it is convienent for my setup.

firesurfer:
@fmauch I just reworked parts of the PRs and tested them all merged together in one branch on our system and it worked just fine. 

As stated before I didnt test them seperately as this would be way to time consuming at the moment

firesurfer:
Closing this PR as we merged almost all the parts

PR #568, opened 2022-12-09
Added method to setRealtimePriority of the controller manager thread

firesurfer:
This PR adds a method to increase the priority of the controller manager thread and tie the thread to a core 1.

This method of achieving near realtime performance  for systems without  the linux RT patches has been successfully tried and is in use in the RSL ethercat stack.

The original code for this method can be found in: https://github.com/leggedrobotics/ethercat_sdk_master/blob/master/src/ethercat_sdk_master/EthercatMaster.cpp

In order for this to work without sudo you will need add the following lines to `etc/security/limits.conf`

```
* soft rtprio 99
* hard rtprio 99
```

I also used the same code succesfully for a setup with two UR16e controlled from the same PC (without ros2control).

### DIFF
--- a/ur_robot_driver/src/ur_ros2_control_node.cpp
+++ b/ur_robot_driver/src/ur_ros2_control_node.cpp
@@ -55,8 +55,21 @@
 
 #include <pthread.h>
 
+/**
+ * @brief set the priority of the current thread to the passed priority and tie the thread to a certain core
+ * @param priority - the desired priority between [0,99] with 99 beeing the highest priority
+ * @param cpu_core - the number of the core you want to tie the thread to (0,your amount of cores]. You are not allowed to tie the thread to core 0 as core 0 is often used for kernel tasks!
+ * 
+ * @note In order for this method to work without root permissions you need to add the following lines to /etc/security/limits.conf (without the ticks "")
+ *    "* soft rtprio 99"
+ *    "* hard rtprio 99"
+ * You need to logout and log back into your session in order for these changes to be applied
+
+*/
 bool setRealtimePriority(int priority, int cpu_core){
   //Taken from https://github.com/leggedrobotics/ethercat_sdk_master/blob/6b420bc1785cf26324aab62c79347b2a6e07924d/src/ethercat_sdk_master/EthercatMaster.cpp#L121
+  
+
   bool success = true;
   //Handle to our thread
   pthread_t thread = pthread_self();
--- a/ur_robot_driver/src/ur_ros2_control_node.cpp
+++ b/ur_robot_driver/src/ur_ros2_control_node.cpp
@@ -38,12 +38,73 @@
 #include <thread>
 #include <memory>
 
+
 // ROS includes
 #include "controller_manager/controller_manager.hpp"
 #include "rclcpp/rclcpp.hpp"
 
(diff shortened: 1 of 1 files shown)


fmauch:
This looks like a great addition, @destogl might also be interested in. In fact, I planned to migrate to [`realtime_tools`](https://github.com/ros-controls/realtime_tools/blob/master/include/realtime_tools/thread_priority.hpp) to handle thread priority. I think having this as a method when no RT kernel is available would be a great addition there, too!

firesurfer:
I am also working on a version of the driver which uses the standard ros2control node. The standard node already uses the realtime_tools package. (But does not have the ability to tie a thread to a certain core) 

From my point of view the only reason to use a custom ros2control node is that the read method in the ur_driver waits per default. I would guess setting `non_blocking_read` to true might actually already do the job

fmauch:
Yes, we deliberately created an own control node as the ros2_control_node uses a sleep at the end of each cycle in order to hold the cycle time. Since in our case we actually **want** the robot to dictate  the control cycle, we have written our own.  See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/281 or https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/210 for the whole discussion.

fmauch:
Thanks again @firesurfer for pushing me into researching this a bit more. We've [run some tests](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/60f08359f6/ur_robot_driver/doc/real_time_benchmarking.md) (on the ROS1 driver, but the results should be applicable to ROS2, as well). And indeed using FIFO scheduling on other kernels seems to have a significant effect.

I cannot see a significant improvement from the thread affinity and configuring this requires knowledge / assumptions about the target system or manual configuration. Therefore, I think it would be preferable to restrict that to using FIFO scheduling.

Also using a non-blocking read seems to help stabilizing cycle times a lot, so it might make sense to migrate back to ros2_control's hardware interface node, once FIFO scheduling on non-RT kernels is also supported.

firesurfer:
@fmauch Wow these are very extensive tests. 
I totaly agree with you regarding the thread -> core affinity. From my experiments in the past I can confirm that this makes only a very small difference. 

So how are we going to progress with the scheduling enhancements / multiple arms support?

fmauch:
I suggest breaking down things a little. Regarding the scheduling enhancements, I would take the detour of integrating this into ros2_control and then switching this driver towards using the default ros2_control node. Regarding the multi-robot stuff, see https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/567#issuecomment-1410619159

firesurfer:
I guess this can be closed now

firesurfer:
@fmauch I did some further research on tying a thread to a certain core (which is the reason why I added the PR to https://github.com/ros-controls/realtime_tools/pull/140. 

It really depends on what kind of system (cpu, etc) you are running. I guess your PC has an older CPU which neither has E/P cores (such as the newer intel processors) or 3D Cache such as some of the newer AMD processors. On such systems swapping between cores on different chiplets /core may be more expensive in terms of cache misses etc.

There is no reason to change anything because of that. Just wanted to follow up on that.

PR #579, opened 2023-01-09
Added service to control force mode from ROS2

urmarp:
Added service to set force mode

Needs https://github.com/ros-industrial/ur_msgs/pull/21 and https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/75

## ToDo list

- [x] Make force_mode work
- [x] Implement an own controller for force_mode
- [ ] Document force_mode_controller
- [ ] Make controller thread_safe using realtime_buffer
- [ ] Test and probably improve force_mode_controller in conjunction with moving the robot
- [ ] Add test(s) for force_mode_controller

### DIFF
--- a/ur_controllers/CMakeLists.txt
+++ b/ur_controllers/CMakeLists.txt
@@ -8,6 +8,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(angles REQUIRED)
 find_package(controller_interface REQUIRED)
+find_package(geometry_msgs REQUIRED)
 find_package(joint_trajectory_controller REQUIRED)
 find_package(lifecycle_msgs REQUIRED)
 find_package(pluginlib REQUIRED)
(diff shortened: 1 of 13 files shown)


fmauch:
Since I did some amount of work on this, I think this should be reviewed by @RobertWilbrandt.

In a nutshell I think it is very discussable whether we add this functionality to the IO controller or whether we should shift it to another controller. As we put a lot of stuff into the IO controller, I don't see why this should not go into this, as well, but I definitively don;t like getting this thing bigger and bigger.

Also, tests would be good, but that would also go for the other components inside that controller.

fmauch:
Thinking about this: I really think, this should go into its own controller. Then activating / deactivating the controller would be activating / deactivating force_mode providing a service for parametrizing the force_mode.

Also, setting force_mode's damping and gain should be handled from this.

fmauch:
Closing this in favor of #1049

PR #587, opened 2023-01-31
Update the toplevel README for humble

fmauch:
The README has been updated on the main branch, also with respect to the binary installation.
This commit ties them together, pointing to humble-specific endpoints,
where necessary.


RobertWilbrandt (approved):
This seems correct. Some of the information (e.g. clang-format-14 instead of 10) is now more up-to-date then the main branch, we should fix that in a next step.

PR #589, opened 2023-02-06
Correct calibration correction launch file in doc

RobertWilbrandt:


fmauch (approved):
Thanks for spotting this!

PR #594, opened 2023-02-08
Ported controllers to generate_parameters libary and added prefix for controllers (Multiarm part 2)

firesurfer:
This PR is part of splitting up #567  into multiple parts 

This PR ports the gpio controller and the speed scaling state broadcaster to generate_parameter_library

Addtionally it applies the prefix specified in the urdf in order to avoid name clashes.

NOTE: If really necessary this work could be split up into two PRs (one for the prefix, one for the  generate_parameter  library) but this would be a lot of work for just a small benefit.


fmauch:
@firesurfer yes, reverting that seems reasonable. Would you like to do that and I'll review it afterwards? Thanks!

firesurfer:
@fmauch I just reverted the behaviour to the default behaviour

ur_controllers/CMakeLists.txt
```diff
@@ -32,19 +33,36 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
...
+generate_parameter_library(
+  speed_scaling_state_broadcaster_parameters
+  src/speed_scaling_state_broadcaster_parameters.yaml
+)
+
+
+
 add_library(${PROJECT_NAME} SHARED
-  src/scaled_joint_trajectory_controller.cpp
+  #src/scaled_joint_trajectory_controller.cpp
```
fmauch, line 55: The scaled JTC should not get removed in this PR.

ur_controllers/src/speed_scaling_state_broadcaster_parameters.yaml
```diff
@@ -0,0 +1,11 @@
+speed_scaling_state_broadcaster:
+   prefix: {
+    type: string,
+    default_value: "",
+    description: "urdf prefix of the corresponding arm"
+   }
+   state_publish_rate: {
+    type: double,
+    default_value: 100.0,
+    description: "rate the speedscaling state will be updated"
```
fmauch, line 10:
```suggestion
    description: "Rate at which the speedscaling state will be published."
```

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -427,7 +447,7 @@ URPositionHardwareInterface::on_activate(const rclcpp_lifecycle::State& previous
                         "README.md] for details.");
   }
 
-  ur_driver_->startRTDECommunication();
+
```
fmauch, line 450: What happened to this? This change seems like it would break everything.

fmauch (changes requested):
As far as I can see, the proposed changes would break a couple of things. We should make sure that everything is running together with this PR applied.

fmauch (approved):
Looks good to me and tested this locally. Thank you @firesurfer for the continuous effort!

fmauch:
@Mergifyio backport humble

PR #595, opened 2023-02-08
Adds full nonblocking readout support (Multiarm part 4)

firesurfer:
Additionally this PR adds fix for shutting down the hardware interface. (Calling deactivate at destruction)

split up of #567


firesurfer:
I just rebased this PR onto #596 otherwise there will be merge conflicts

PR #598, opened 2023-02-28
Fix cmake dependency on controller_manager

RobertWilbrandt:
The buildfarm currently fails due to an include error on `controller_manager` (see [here](https://build.ros2.org/job/Rbin_ujv8_uJv8__ur_robot_driver__ubuntu_jammy_arm64__binary/89/console)) and i suspect this will fix that.


RobertWilbrandt:
Might be that this is not even needed, see https://github.com/ros-planning/moveit2/issues/1840. It might still make sense to merge this though.

stefanscherzinger (approved):
Yes, thanks. That's cleaner and should indeed fix the node's missing dependency on `controller_manager`.

PR #603, opened 2023-03-13
Perform scheduled runs of the humble source build

RobertWilbrandt:
We also run rolling in scheduled builds, so this is more consistent. I find this to be useful, as source builds especially can show errors from different packages and our ci doesn't reflect fixes to those until we push something new to the corresponding branch.

### DIFF
--- a/.github/workflows/humble-source-build.yml
+++ b/.github/workflows/humble-source-build.yml
@@ -1,0 +1,41 @@
+name: Humble Source Build
+on:
+  schedule:
+    # Run every morning to detect flakiness and broken dependencies
+    - cron: '43 1 * * *'
+
+jobs:
+  humble_source:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    container:
+      image: ubuntu:jammy
+    env:
+      ROS_DISTRO: humble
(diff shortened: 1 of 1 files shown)


.github/workflows/humble-source-build.yml
```diff
@@ -0,0 +1,41 @@
+name: Humble Source Build
+on:
+  schedule:
+    # Run every morning to detect flakiness and broken dependencies
+    - cron: '43 1 * * *'
+
+jobs:
+  humble_source:
+    runs-on: ubuntu-20.04
```
stefanscherzinger, line 9: Could that also be

```suggestion
    runs-on: ubuntu-22.04
```
or is that somehow less stable as host OS?

stefanscherzinger (approved):
Yes, seems a good idea.

.github/workflows/humble-source-build.yml
RobertWilbrandt: It does not really matter as everything relevant is run inside of a container anyway. We could change it, but i would only do that if we change the corresponding workflow(s) on the humble branch at the same time to avoid inconsistencies and i don't see any reason to do so now.

PR #605, opened 2023-03-13
Increase timeout for first test service call to driver

RobertWilbrandt:
Previously, only the dashboard client was allowed to wait longer for the first service call, leading to flaky tests in cases where RTDE could not directly initialize. This can be seen in e.g. [this CI run](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/runs/4401593654/jobs/7707936503).

### DIFF
--- a/ur_robot_driver/test/robot_driver.py
+++ b/ur_robot_driver/test/robot_driver.py
@@ -147,10 +147,17 @@ class RobotDriverTest(unittest.TestCase):
 
     def init_robot(self):
 
-        # Wait longer for the first service client as the robot driver is still starting up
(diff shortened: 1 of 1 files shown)


stefanscherzinger (approved):
This looks good to me. :+1: 

At some later point, it might be worthwhile to fix *Failed to initialize RTDE client, retrying in 10 seconds* in the first place.

RobertWilbrandt:
I am not so sure if we need to look into the RTDE timeout - I think this is just an artifact of ursim booting and the dashboard server being available earlier than RTDE. This seems like the correct behavior to me.

PR #608, opened 2023-03-14
Controller spawner timeout

RobertWilbrandt:
After merging #605 the controller spawner timeout quite often leads to CI failures: If the initial RTDE connection fails, the driver waits 10 seconds for a retry, leading to timeouts in the controller spawners. This PR fixes that problem by allowing a longer timeout for controller spawners.


stefanscherzinger (approved):
I could follow the idea of the new timeout and it seems reasonable. Nice usage with the `controller_spawners`.

PR #614, opened 2023-03-20
Fix humble source CI through `ros2_tracing`

stefanscherzinger:
## Problem
We get `/opt/ros/rolling/lib/libcontroller_manager.so: undefined reference to tracetools::detail::get_symbol_funcptr(void*)`
in our CI on *Rolling* from source.

## Workaround
It seems that `ros2_tracing` recently changed their function signature from

`const char * get_symbol_funcptr(void * funcptr)` to
`char * get_symbol_funcptr(const void * funcptr)`
in [faf6c27](https://github.com/ros2/ros2_tracing/commit/faf6c278faa23ab2f66ecd6e8fcc674138193a20).

For some reason, the dependent packages don't seem to have updated yet.

Let's see if the previous version fixes the from-source build.

### DIFF
--- a/Universal_Robots_ROS2_Driver-not-released.rolling.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.rolling.repos
@@ -4,6 +4,10 @@
 # Once Upstream packages are released and synced to the target distributions in the required
 # version, the entry in this file shall be removed again.
 repositories:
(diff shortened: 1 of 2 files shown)


fmauch:
Since we are planning to drop the source builds I'd close this PR.

PR #616, opened 2023-03-22
FollowJointTrajecotry Test Configuration

livanov93:
# Description
- PR depends on https://github.com/eholum/ros2_controllers/pull/1 which also needs to be moved to the main repo
- PR introduces launch and configuration files for testing action interface for JTC

### DIFF
--- a/ur_robot_driver/config/test_action_goals_config.yaml
+++ b/ur_robot_driver/config/test_action_goals_config.yaml
@@ -1,0 +1,28 @@
+action_client_position_trajectory_controller:
+  ros__parameters:
+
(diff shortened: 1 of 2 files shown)


RobertWilbrandt:
I don't understand what exactly this depends on - is there a corresponding `ros2_controllers` PR upstream? The test also needs to be registered in the CMakeLists.txt and start its own ursim, is that part still missing?

urfeex:
@livanov93 I would go ahead and close this. This seems to be depending on another suggestion that never got merged, right? Feel free to reopen this if it comes relevant, again.

PR #618, opened 2023-04-04
Added ur_moveit_config.launch.py arguments for publishing robot description and robot description semantic

haavarpb:
Hi,

I am working with a project where I use the MoveGroupInterface for moving the Universal Robot. To use it you either have to load the robot description and the robot description semantic (as you already do in the ur_moveit_config.launch.py) into the node that will utilize the MoveGroupInterface, or you can pull it from the move_group automatically. The latter is better in my view but for this to work I need to instruct the move_group node to publish them. This would make it easier for people to use MoveGroupInterface with your driver. 

Relevant resources: 
https://github.com/ros-planning/moveit2_tutorials/issues/525
https://moveit.picknik.ai/humble/doc/examples/urdf_srdf/urdf_srdf_tutorial.html

This solution should work also in other branches other than galactic. For galactic to build tho, I had to update the .repos file setting the UR Description package to galactic instead of ros2.

### DIFF
--- a/Universal_Robots_ROS2_Driver.galactic.repos
+++ b/Universal_Robots_ROS2_Driver.galactic.repos
@@ -6,7 +6,7 @@ repositories:
   Universal_Robots_ROS2_Description:
     type: git
     url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
-    version: ros2
+    version: galactic
   ur_msgs:
     type: git
     url: https://github.com/ros-industrial/ur_msgs.git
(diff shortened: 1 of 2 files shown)


fmauch:
Thank you for providing this @haavarpb. Since Galactic is EOL by now we're not going to add any new features to the galactic branch, though. Would you like to make this PR to main, as well? Otherwise I'll do that myself.

haavarpb:
Hehe, I get you, no worries. I am just still stuck in the past for a little while longer ;-D

I can open a pull request for main also :-)

PR #622, opened 2023-04-06
Debug ci failures

RobertWilbrandt:
This should increase insights into CI failures


fmauch (approved):
This should help in making the test output more readable.

PR #623, opened 2023-04-06
Fix ros2 control repos

RobertWilbrandt:


fmauch (approved):
yes, this absolutely makes sense, since we build the direct dependencies in the semi-binary builds.

PR #624, opened 2023-04-06
Add ros2_control as direct dependency

RobertWilbrandt:
Only directly using ros2_controllers in semi-binary builds is prone to breakage


fmauch (approved):
makes sense

PR #626, opened 2023-04-11
Do not require building moveit2 from source

fmauch:
As MoveIt2 is not a direct dependency and released upstream, we can remove it again.

This was introduced in #610 but got forgotten to revert in #621.


RobertWilbrandt (approved):
We could start discussing if moveit isn't also an "immediate dependency", but as long as we don't test its execution in CI i agree that this should be removed.

PR #627, opened 2023-04-12
Remove source builds from CI

fmauch:
The source builds did not really seem to be useful in the past. They fail a lot of the times and that is mostly due to some breaks somewhere up the dependency chain that does not really affect us.

We already have multiple mechanisms in place to detect incompatibilities early (testing repo, semi-binary builds) so that we think we can live very well without the source builds. In the end less day-to-day failing of pipelines increases the awareness if a pipeline really fails for a relevant reason.


README.md
```diff
@@ -41,10 +41,6 @@ Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) a
          <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/foxy-semi-binary-build.yml/badge.svg?event=schedule"
               alt="Foxy Semi-Binary Build"/>
       </a> <br />
```
RobertWilbrandt, line 43:
```Suggestion
      </a>
```

I don't think these breaks belong here anymore (for the other entries as well)

RobertWilbrandt (changes requested):
I completely agree with the reasoning, i don't know any example on top of my head where we saw an error in source builds that were not either visible in the other tests as well or in some remote third-degree dependency.

README.md
fmauch: Fixed in 41594d3

RobertWilbrandt (approved):

PR #593, opened 2023-02-08
Adding additional parameters needed as step towards multiarm support (Multiarm part 1)

firesurfer:
This PR is part of splitting up into multiple parts #567 

Note:  The non-blocking read won't properly work like this without the PR in this series that allows starting the rtde communication at the latest point of time possible

PR #633, opened 2023-04-17
Adding additional parameters needed as step towards multiarm support (Multiarm part 1, recreation of #593)

fmauch:
This PR is part of splitting up into multiple parts https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/567

Note: The non-blocking read won't properly work like this without the PR in this series that allows starting the rtde communication at the latest point of time possible

### DIFF
--- a/Universal_Robots_ROS2_Driver-not-released.humble.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.humble.repos
@@ -4,3 +4,7 @@
 # Once Upstream packages are released and synced to the target distributions in the required
 # version, the entry in this file shall be removed again.
 repositories:
+  Universal_Robots_ROS2_Description:
+    type: git
+    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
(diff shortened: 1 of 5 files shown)


fmauch:
I'll squash this manually to make sure the commit is correctly attributed to @firesurfer.

destogl (approved):
Looking good.

fmauch:
I just added missing repo files updates in https://github.com/firesurfer/Universal_Robots_ROS2_Driver/pull/1 Waiting or this before merging.

firesurfer:
@fmauch I see. Shall we implement a workaround until this issue is solved? Perhaps passing `~` per default and handling that value in the hardware interface?

fmauch:
Yes, as a workaround we could add a leading character that we remove inside the hardware interface. Not nice, but that could help us progressing, at least.

fmauch:
I just tested this locally against the current version of the description and it seems to work fine -> merging

PR #636, opened 2023-04-21
Use individual workflows for main and testing on rolling

fmauch:
As suggested by @destogl in the past this will give us individual badges for main and testing while keeping maintenance efforts low.


fmauch:
I only updated humble and rolling. For galactic things are being kept as long as they work, for foxy I would not expect any upstream package to fail so it does not make much sense to differentiate between main and testing. Furthermore, the sim launch inside the tests was never ported back to foxy, so we can't use the same template there.

destogl (approved):
It looks good to me. Just restart CI to see why the error. I think it is because many things are running at once. It happens from time to time.

README.md
```diff
@@ -53,23 +53,39 @@ Check also [presentations and videos](ur_robot_driver/doc/resources/README.md) a
...
-              alt="Humble Binary Build"/>
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-binary-main.yml?query=event%3Aschedule++">
+         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-binary-main.yml/badge.svg?event=schedule"
+              alt="Humble Binary Main"/>
       </a> <br />
-      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-semi-binary-build.yml?query=event%3Aschedule++">
-         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-semi-binary-build.yml/badge.svg?event=schedule"
-              alt="Humble Semi-Binary Build"/>
+      <a href="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-binary-testing.yml?query=event%3Aschedule++">
+         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-binary-main.yml/badge.svg?event=schedule"
```
RobertWilbrandt, line 61:
```Suggestion
         <img src="https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/workflows/humble-binary-testing.yml/badge.svg?event=schedule"
```

PR #639, opened 2023-04-24
Load joint limits in MoveIt launch file

sea-bass:
This PR accompanies https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/62, and ensures that the joint limits files for each specific robot config are actually passed into the example MoveIt launch file.


RobertWilbrandt:
@sea-bass i opened an alternative PR in #645, could you check if that would also work for you?

PR #640, opened 2023-04-25
Added workflows for iron

fmauch:
This PR adds CI builds for Iron. The idea is that we create an iron branch once this is merged onto main and we keep the iron branch up-to-date with main as long as rolling is compatible with iron.


fmauch:
Closing in favor of #709

PR #645, opened 2023-04-25
Define default maximum accelerations for MoveIt

RobertWilbrandt:
This defines default values for maximum joint accelerations, as MoveIt is otherwise not able to perform time parametrization. The 5rad/s² are from the discussion in https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/issues/46.

This patch only passes these limits to MoveIt. While we could also add them in `ur_description`, this seems incorrect. The robot does not inherently have any limits on joint accelerations (only on torques), and other software using `ur_description` might be able to handle this differently.

### DIFF
--- a/ur_moveit_config/config/joint_limits.yaml
+++ b/ur_moveit_config/config/joint_limits.yaml
@@ -1,0 +1,25 @@
+# These limits are used by MoveIt and augment/override the definitions in ur_description.
+#
+# While the robot does not inherently have any limits on joint accelerations (only on torques),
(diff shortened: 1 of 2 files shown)


sea-bass (approved):
Tested locally and it works for me!

fmauch (approved):
Seems good to me and seems to be working well. Tested against MoveIt! 2.7.1.

fmauch:
Thanks for testing this, as well @sea-bass

PR #648, opened 2023-04-25
Wait longer for controllers to load and activate

RobertWilbrandt:
This came up in the backport of CI fixes https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/647 and seemed to be unrelated to those changes.


fmauch (approved):
Maybe we should think about querying the controller manager for all required controllers being activated, but for now this seems reasonable and important.

PR #658, opened 2023-04-28
Add ur_moveit.launch.py arguments to control whether the move_group node publishes robot description and robot semantic description

haavarpb:
Users can now pass arguments to the launch file to make the move_group node publish descriptions needed for using the MoveGroupInterface.

Until now, if you want to use the MoveGroupInterface in a custom node, you'd have to copy the entire ur_moveit.launch.py to load and construct the urdf and srdf information needed for the move_group node and the MoveGroupInterface node. This results in a rather huge launch file for the single MoveGroupInterface-node to work.

This PR is a quality of life change so that other people wanting to use MoveGroupInterface with the UR can do so by including the existing ur_moveit.launch.py in their own file using IncludeLaunchDescription and then launch their MoveGroupInterface on top of that which pulls the required urdf and srdf from the move_group-node.

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -60,6 +60,8 @@ def launch_setup(context, *args, **kwargs):
     use_sim_time = LaunchConfiguration("use_sim_time")
     launch_rviz = LaunchConfiguration("launch_rviz")
     launch_servo = LaunchConfiguration("launch_servo")
+    publish_robot_description = LaunchConfiguration("publish_robot_description")
+    publish_robot_description_semantic = LaunchConfiguration("publish_robot_description_semantic")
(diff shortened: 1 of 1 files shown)


RobertWilbrandt:
Publishing the `robot_description_semantic` definitely seems useful to me. I don't quite get why MoveIt needs to publish the `robot_description` though - are there use cases where this is not done by the `robot_state_publisher`?

haavarpb:
Good question 

> Publishing the `robot_description_semantic` definitely seems useful to me. I don't quite get why MoveIt needs to publish the `robot_description` though - are there use cases where this is not done by the `robot_state_publisher`?

Good question. This could possibly be a good suggestion to make on the MoveIt repo!

fmauch:
Closing this in favor of 1036

Issue #656, opened 2023-04-26
Scaled JTC seems to segfault with latest ros2_controllers


fmauch:
Our semi-binary pipeline for humble seems to be segfaulting in the `update` function from the scaled JTC. The commit after this started to happen seems to be https://github.com/ros-controls/ros2_controllers/pull/579, but at the moment we are not sure whether this is related or not.

fmauch:
Source of the problem seems to be, that the jtc configures [exposes its headers as `PRIVATE`](https://github.com/ros-controls/ros2_controllers/blob/8e6697b8e9509265d8c32a9331f537b6af8cd1ed/joint_trajectory_controller/CMakeLists.txt#L43).

Since it also installs the headers, this will work once the headers got installed into `/opt/ros/humble/include`, but not for a local workspace. The package's include directory will not get added to the list of `ur_controller`'s include directories. Therefore, the version from `/opt/ros/humble/include` will be used when building our scaled controller. However, runtime linking will use the `libjoint_trajectory_controller.so` from our workspace which is not ABI compatible. There's not much we can do to mitigate this, since the JTC package will be installed as part of MoveIt`s dependency chain. One way to fix this would be to fix include exports in the JTC similar to https://github.com/ros-controls/ros2_controllers/pull/510.

I expect that this will "fix" itself, once a ros2_controllers version with those changes will be released. Nevertheless, this raises the necessity of tackling #660.

fmauch:
This is now not  a problem anymore, hence closing this.

PR #659, opened 2023-04-28
Attempt to fix semi-binary build: Do not install binary version of JTC

fmauch:
This PR attempts to workaround #656.


fmauch:
Since the JTC package gets installed as a MoveIt!-dependency that approach doesn't make much sense.

PR #666, opened 2023-05-08
Remove ur_bringup package

fmauch:
We've added deprecation notices for this quite a while ago. With Iron around the corner it is time to actually remove the package.


RobertWilbrandt (approved):
Seems complete to me

PR #673, opened 2023-05-15
Adds full nonblocking readout support (Multiarm part 4)  - v2

firesurfer:
Recreated #595


firesurfer:
Just tested this PR on my setup with two UR16e. Worked fine so far

fmauch:
@firesurfer since the changes from this PR are rather small, I think it would be beneficial (and should be fairly easy to do) to break it up into two commits:

 * adding `on_deactivate` in the destructor 
 * moving `startRTDECommunication` (+adding the flag)

Thanks again for sticking with this.

firesurfer:
I removed the on_deactivate in the destructor and will add an additional PR for that.

firesurfer:
Can you please test on your setup if the driver also has a connection drop when zeroing the FT Sensor @fmauch 

I am rather clueless why this should happen given that this doesn't happen on my branch where I test merged all commits in like february https://github.com/firesurfer/Universal_Robots_ROS2_Driver/tree/merge_test

firesurfer:
Interesting. Then lets merge this and I add a new issue for the zeroing

fmauch:
I already did in #686

fmauch:
Just tested this against current main (with and without headless mode) and I don't see the problems you mentioned.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -514,6 +513,12 @@ void URPositionHardwareInterface::asyncThread()
 hardware_interface::return_type URPositionHardwareInterface::read(const rclcpp::Time& time,
                                                                   const rclcpp::Duration& period)
 {
+  // We want to start the rtde comm the latest point possible due to the delay times arising from setting up the
+  // communication with multiple arms
+  if (!rtde_comm_has_been_started_) {
+    rtde_comm_has_been_started_ = true;
+    ur_driver_->startRTDECommunication();
```
fmauch, line 520: Logically, I believe it makes sense to swap those lines.
```suggestion
    ur_driver_->startRTDECommunication();
    rtde_comm_has_been_started_ = true;
```

fmauch (approved):
Looks good to me.

fmauch:
@Mergifyio backport humble

PR #679, opened 2023-05-24
Calling on_deactivate in dtr

firesurfer:
See title


fmauch (approved):
This seems like a fairly good idea to do.

In general, I think this is something that could / should be handled inside the base classes of ros2_control, but having it here until this happens definitively won't hurt.

fmauch:
@Mergifyio backport humble

Issue #674, opened 2023-05-15
scaled JTC does not build against humble version of ros2_controllers


firesurfer:
```
scaled_joint_trajectory_controller.cpp:274:16: error: no matching function for call to ‘ur_controllers::ScaledJointTrajectoryController::publish_state(const rclcpp::Time&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&)’
  274 |   publish_state(time, state_desired, state_current, state_error);
      |   ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Apparently the definition of publish_state in the JTC has changed.

WaReStu:
Okay, this seems to be exactly the same problem I encounter during a build from source as per the main page ([main-page](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver#build-from-source)).

Just to save others with really fresh installs from searching (like I did): Right now, neither the binary install of the Driver, nor the build from source work successfully for **me** on two seperate ubuntu 22.04 desktops.

Now keeping to the source build: This happens to me during Step 4:
`colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release`

> --- stderr: ur_controllers
> /home/tallarean/ws_ur/ros_ur_driver/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp: In member function ‘virtual controller_interface::return_type ur_controllers::ScaledJointTrajectoryController::update(const rclcpp::Time&, const rclcpp::Duration&)’:
> /home/tallarean/ws_ur/ros_ur_driver/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:274:16: error: no matching function for call to ‘ur_controllers::ScaledJointTrajectoryController::publish_state(const rclcpp::Time&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&, joint_trajectory_controller::JointTrajectoryController::JointTrajectoryPoint&)’
>   274 |   publish_state(time, state_desired, state_current, state_error);
[...]

PR #682, opened 2023-05-24
added nasty conditional compilation for compiling scaled JTC on humble

firesurfer:
This is definitely not a nice solution but it is a solution which allows to compile the current main branch on humble. 

Different solutions that I tried were:

1. Pass in ROS_DISTRO environment varible as define -> Preprocessor cant do string compares
2.  if constexpr -> Both code paths have to be valid 
         -> With SFINAE one could hack this together...

A different solution could be to ask the JTC package maintainers to include the ament_generate_version_header statement. This will generate a header file with the  version numbers defined in the package.xml which one could check directly.

Other than that..perhaps something type erasure like where the compiler can't check the method signature anymore :D


firesurfer:
@fmauch An alternative would be to use the solution from the cartesian motion controllers: https://github.com/fzi-forschungszentrum-informatik/cartesian_controllers/blob/1239413138acf2bf16716c38337a9e703602ba03/cartesian_controller_base/CMakeLists.txt#L11

firesurfer:
@fmauch Are you in general interested in such a solution? Implementing something like this would allow building main against humble I think

fmauch:
As  written in #674 it is not intended to build the main branch on humble, at all. The main branch is for the most recent changes that are meant to run with rolling. We try to port back as much as possible to humble and iron. If there is something missing in your opinion, please open a separate issue about this.

firesurfer:
@fmauch Just to comment on this (I do not want to open an additional issue given that there is nothing missing)

I used the main branch in combination with humble for quite some time now and the scaled JTC was the only thing that was causing issues there. I guess you are right, if there are more breaking changes regarding ros2(control) in the future this might be unfeasible to maintain but for now I don't see any reason why one wouldn't want to build the main branch on humble because it works.

PR #688, opened 2023-05-24
Use tf prefix properly

fmauch:
Since https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/66 is merged, we'll have to revert the workaround in the driver as well.

PR #690, opened 2023-05-24
Also build control_msgs from source for now

fmauch:
In #688 we built ros2_control from sources. However, we also must build the control_msgs from source, as well.


fmauch:
This indeed fixes the build for rolling main. @RobertWilbrandt could you please review?

RobertWilbrandt (approved):

PR #707, opened 2023-06-05
Added services to set tool voltage and zero force torque sensor - galactic backport (#466)

srsidd:
Added launch arguments for reverse ip and script command interface port.
Tested with hardware (UR5) and worked without error when set in remote control mode.

Backport to galactic. Goes with https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/73.

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -1,0 +1,560 @@
+# Copyright (c) 2021 PickNik, Inc.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#
+#    * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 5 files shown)


ur_robot_driver/launch/ur_control.launch.py
```diff
@@ -0,0 +1,560 @@
+# Copyright (c) 2021 PickNik, Inc.
```
fmauch, line 1: Since we don't have launchfiles in `ur_robot_driver` on galactic, I guess it doesn't make much sense to add this here.

fmauch (changes requested):
Thank you for providing this. I guess, you are aware, that galactic is EOL since a while, hence there won't be any support on this, anymore. However, since you took the effort of porting this and from a quick glance this seems fine, I would be merging this anyway. However, as noted, I think the additional launchfile should be removed.

---

Edit: Now seeing that there is an accompanying MR on the description, I am against putting that effort into galactic. I would rather suggest switching to Humble.

fmauch:
Closing this as galactic has been EOL for quite some time now.

PR #708, opened 2023-06-05
Use binary released packages for binary builds only.

fmauch:
The idea behind this is basically this:
 * If we require direct upstream dependencies in a more recent version, we have the semi-binary builds. With this, it would be acceptable, that the binary builds turn failed if there has been an upstream breaking change, that we already adapted to.
 * For development the semi-binary builds are the actually interesting ones. They will keep building upstream dependencies from their most recent versions.
 * The binary builds should give us feedback whether the driver will work with the currently released packages. This is only a valid feedback if it is actually built against the released state.

This way, situations like it just happened that our humble release was failing should pop up in the binary-testing earlier in the process, so that we can react on this. If we build our binary builds against upstream master, this migh get (and was) hidden.

### DIFF
--- a/Universal_Robots_ROS2_Driver-not-released.humble.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.humble.repos
@@ -4,19 +4,3 @@
 # Once Upstream packages are released and synced to the target distributions in the required
 # version, the entry in this file shall be removed again.
 repositories:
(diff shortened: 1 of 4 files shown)


RobertWilbrandt (approved):
This seems like a good to step to improve CI usability for our development

Issue #697, opened 2023-05-26
Apply fix for trajectory execution race condition to ScaledJointTrajectoryController


schornakj:
The JointTrajectoryController in ros2_controllers had [a race condition](https://github.com/ros-controls/ros2_controllers/pull/565) in which trajectory execution could erroneously succeed while the robot was still in motion, which caused undesirable behavior when performing multiple motion plans and trajectories in series.

Since this repo's ScaledJointTrajectoryController is based on JointTrajectoryController, it probably also has the same problem.

To resolve this, the changes from https://github.com/ros-controls/ros2_controllers/pull/565 need to be applied to ScaledJointTrajectoryController.

panagelak:
@schornakj  @fmauch 

I create a pr for this for humble branch https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/710

fmauch:
Should be resolved by #810, #811, #813

PR #710, opened 2023-06-05
Fix SJTC from immediately returning success (#697)

panagelak:
I applied the fix (for humble) which was done at the ros2_controllers repo https://github.com/ros-controls/ros2_controllers/pull/565 at the commit 2f546066827cf90020d6040c180ae564407b9bae

This pr refers to your newly opened issue #697 

I hope it is merged : ) cheers

### DIFF
--- a/ur_controllers/src/scaled_joint_trajectory_controller.cpp
+++ b/ur_controllers/src/scaled_joint_trajectory_controller.cpp
@@ -115,6 +115,8 @@ controller_interface::return_type ScaledJointTrajectoryController::update(const 
     fill_partial_goal(*new_external_msg);
     sort_to_local_joint_order(*new_external_msg);
     traj_external_point_ptr_->update(*new_external_msg);
+    // set the active trajectory pointer to the new goal
+    traj_point_active_ptr_ = &traj_external_point_ptr_;
   }
 
   JointTrajectoryPoint state_current, state_desired, state_error;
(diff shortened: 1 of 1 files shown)


ur_controllers/src/scaled_joint_trajectory_controller.cpp
```diff
@@ -262,6 +267,7 @@ controller_interface::return_type ScaledJointTrajectoryController::update(const
               result->set__error_code(FollowJTrajAction::Result::GOAL_TOLERANCE_VIOLATED);
               active_goal->setAborted(result);
               rt_active_goal_.writeFromNonRT(RealtimeGoalHandlePtr());
+
```
schornakj, line 270: Would we need to also set `traj_point_active_ptr_ = nullptr;` at this line, since we reset the goal pointer here?

fmauch:
Thank you for providing this. Me or @RobertWilbrandt will have to look into this more deeply.

ur_controllers/src/scaled_joint_trajectory_controller.cpp
panagelak: yes i thought about that too, but i did not see it in the original too when return error code goal_tolerance_violated.
this could be an error on the original or not needed or potentially breaking if inserted.
Furthermore it set's the nullptr on_cleanup and on_reset methods which i think are inhereted into the SJTC

fmauch:
I think this should be solved by #811

fmauch:
Thank you @schornakj for bringing that up. Since we've updated the SJTC with the newest version from upstream's JTC this fix is included now (At the time of writing pending for Humble). I'm going to close this therefore.

PR #719, opened 2023-06-20
Use SCHED_FIFO for controller_manager's main thread

fmauch:
[Previous investigations](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/60f08359f6484d1671e12bab6e01fc49eb0fadc8/ur_robot_driver/doc/real_time_benchmarking.md) showed that using FIFO scheduling helps keeping cycle times also non non-RT kernels. This combined with non-blocking read can result in a very stable system.

This is, in fact, very close to what the actual controller_manager_node does except that we always use FIFO scheduling independent of the actual kernel in use.

### DIFF
--- a/ur_robot_driver/src/ur_ros2_control_node.cpp
+++ b/ur_robot_driver/src/ur_ros2_control_node.cpp
@@ -41,6 +41,7 @@
 // ROS includes
 #include "controller_manager/controller_manager.hpp"
 #include "rclcpp/rclcpp.hpp"
+#include "realtime_tools/thread_priority.hpp"
 
 // code is inspired by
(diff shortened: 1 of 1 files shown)


RobertWilbrandt (approved):
This looks good to me. With most of the code the same as `ros2_control_node` from `controller_manager`, we should maybe look at some point if we really need our own controller manager node.

fmauch:
> This looks good to me. With most of the code the same as `ros2_control_node` from `controller_manager`, we should maybe look at some point if we really need our own controller manager node.

Yes, I'll try to push FIFO scheduling into controller_manager, but I wanted to have it running here, before.

fmauch:
@Mergifyio backport humble

Issue #387, opened 2022-06-08
URScript commands for gripper


Abelpy1:
Hi,

In `Universal_Robots_ROS_Driver`, we could send URScript commands to control the Onrobot rg2 gripper on the UR5e, using the topic `/ur_hardware_interface/script_command`. I am very confused that whether we could use URScript commands to control the gripper based on `Universal_Robots_ROS2_Driver (galactic)`?
    
Hope for the reply, thanks.

fmauch:
Closing in favor of #251

Issue #251, opened 2021-12-02
Add support for send custom script commands to robot  


urrsk:
As in the ROS1 driver using the client_library

sillkjc:
I'd like to upvote this as a super important feature I really need. There are many thing that need to be done via string substituted ur scripts that then get sent to the robot. It can also greatly simplify rapid testing / prototyping.

bmdyrdal:
Is there any update on this?

urrsk:
@bmdyrdal to my understanding, ros2_control are still limiting us for doing this. @destogl please correct me if I am wrong
But I agree that this is a need to have feature

bmdyrdal:
@urrsk Thanks for the update. Any recommended ways for working around this?

urrsk:
@bmdyrdal you can modify the UR script the ROS driver/Client library are using and in this way you can adjust it to your needs.
See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/dfa1410653ebcb57f3522491f61a4d6344425056/ur_bringup/launch/ur_control.launch.py#L84

What would you like to do?

bmdyrdal:
@urrsk Aha, I see. Thank you! 
We currently use it during initialization to position the robot correctly and set some output values.
As we also needed to read and write registers we've ended up porting our fork of ur_rtde to ROS2 instead.

sillkjc:
We've found one semi workaround is to write some scripts that SCP a .URP file and .script file to the robot, with the URP being an empty .script command. We can then use the load_program service to load and run it.

This comes with one crippling drawback - load_program also calls load_installation, so the installation from the URP is loaded. 

This resets the currently set tcp/cog/mass safetly limits, home position etc, for each script sent. This makes it not really an option for real hardware.

Still hoping for a 'run_script' service, we added this to the ROS1 UR driver and it proved super useful. This really should have been added before most other services, since you can achieve anything required using scripts.

urrsk:
@sillkjc If you want the robot to only run your own URScript code, then you do not need Polyscope or ROS.

The robot controller accept receiving URScript directly and can received it through a ascii socket connection, though you need to set the Robot in Remote control mode.
(Do not be surprised then you connect the the primary or secondary protocol, as there is a lot of data send from the robot through this port as well.)  
[...]

PR #721, opened 2023-06-20
Urscript interface

fmauch:
This PR adds a bare URScript interface using the `ur_client_library`. This should implement #387 and #251.

In this first version I did a 1to1 sending of the script code towards the robot. If the script code contains an error, it will simply not get executed.

This could get extended in the future, but that should be an easy non-intrusive extension that gets the job done in a lot of situations.

### DIFF
--- a/ur_robot_driver/src/urscript_interface.cpp
+++ b/ur_robot_driver/src/urscript_interface.cpp
@@ -77,7 +77,6 @@ public:
     size_t written;
     m_secondary_stream->write(data, len, written);
   }
-  ~URScriptInterface() override = default;
 
 private:
   rclcpp::Subscription<std_msgs::msg::String>::SharedPtr m_script_sub;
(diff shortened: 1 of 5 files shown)


urrsk:
@fmauch Great work. 

- [ ] Could we add a simple test as well?

fmauch:
@urrsk I've added an integration test for the script interface.

fmauch:
OK, those tests seem not to be reliable. I'll have to do some more work.

ur_robot_driver/src/urscript_interface.cpp
```diff
@@ -0,0 +1,93 @@
...
+        this->get_parameter("robot_ip").as_string(), urcl::primary_interface::UR_SECONDARY_PORT);
+    m_secondary_stream->connect();
+
+    auto program_with_newline = std::string("textmsg(\"urscript_interface connected\")\n");
+    size_t len = program_with_newline.size();
+    const auto* data = reinterpret_cast<const uint8_t*>(program_with_newline.c_str());
+    size_t written;
+    m_secondary_stream->write(data, len, written);
+  }
+  ~URScriptInterface() override = default;
```
RobertWilbrandt, line 80: This seems weird to me, why is it needed?

fmauch: It's not necessarily needed, it's just to explicitly provide an implementation. We can remove that if you prefer.

RobertWilbrandt (approved):

PR #729, opened 2023-06-29
Run robot driver test also with tf_prefix

### LINKED ISSUE #726: Add test using tf_prefix
We should add an integration test starting the driver with a `tf_prefix`. Problems due to a missing backport were showing in #677.

fmauch:
This adds a test (parametrizes the existing test) that also runs the driver with a tf_prefix. This closes #726.

This also serves as an example for a modified controller configuration when using a tf_prefix.


RobertWilbrandt:
I really like automatically testing this, but would prefer to not copy the controller config for it. I think we should be able to use `launch_ros.parameter_descriptions.ParameterFile` with [allow_subst=True](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/parameter_descriptions.py#L180) to to avoid this, i will add that to this PR.

RobertWilbrandt (approved):

RobertWilbrandt:
@fmauch can you take a look if this change seems ok to you?

fmauch:
Thank you @RobertWilbrandt that's a very good addition!

fmauch:
@Mergifyio backport iron

PR #732, opened 2023-07-05
Move communication setup to on_configure instead of on_activate

fmauch:
Corrects point of initialization / connection to the robot by moving it to `on_configure` instead of `on_activate`.

@destogl we talked about this yesterday, I found my commit I had in mind :-)

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/hardware_interface.hpp
+++ b/ur_robot_driver/include/ur_robot_driver/hardware_interface.hpp
@@ -98,7 +98,7 @@ public:
 
   hardware_interface::CallbackReturn on_configure(const rclcpp_lifecycle::State& previous_state) final;
   hardware_interface::CallbackReturn on_activate(const rclcpp_lifecycle::State& previous_state) final;
-  hardware_interface::CallbackReturn on_deactivate(const rclcpp_lifecycle::State& previous_state) final;
+  hardware_interface::CallbackReturn on_cleanup(const rclcpp_lifecycle::State& previous_state) final;
 
   hardware_interface::return_type read(const rclcpp::Time& time, const rclcpp::Duration& period) final;
   hardware_interface::return_type write(const rclcpp::Time& time, const rclcpp::Duration& period) final;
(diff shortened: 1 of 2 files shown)


RobertWilbrandt (changes requested):
This change seems reasonable, but if we start all communication in `on_configure` i would expect the corresponding cleanup to move to `on_cleanup`.

firesurfer:
@fmauch Just saw this PR and thought I take a look at it as it might has the potential breaking multiarm support. Given that `on_configure` and `on_activate` run in different threads (configure in the non rt and activate in the rt one) - If I understood the documentation right - I would suggest changing:

https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/a8a96dc451b5ad52ab61cfc4974d1be1be62b92b/ur_robot_driver/include/ur_robot_driver/hardware_interface.hpp#L227

```
bool rtde_comm_has_been_started_ = false;
```

to

```
std::atomic_bool rtde_comm_has_been_started_ = false;
```


This goes btw. probably also for all other variables which are used from multiple threads (e.g. everything being set by `URPositionHardwareInterface::checkAsyncIO()`)

firesurfer:
@fmauch I agree with you. For passing information between the non-rt and rt thread and `std::atomic_bool` should be used in this case. 


> This goes btw. probably also for all other variables which are used from multiple threads (e.g. everything being set by URPositionHardwareInterface::checkAsyncIO())

In general you are right. Often one can get away with it if only one thread writes and one thread reads and it is okey to have partially updated data.
But in this case I would use the RTBuffer from the realtime_tools package.

fmauch:
@firesurfer yes, thank you for your comments. Thread safety is definitively something we need to tackle at some point. For now, for this PR I would restrict things to the `rtde_comm_has_been_started_` member.

fmauch:
I just rebased and tested this. Especially with the deprecation and [upcoming removal](https://github.com/ros-controls/ros2_control/pull/1237) of passing the robot_description as a parameter and instead using a topic this becomes a required step.

VinDp (approved):
Looks good to me.

fmauch:
@RobertWilbrandt a review from you would be required to merge this.

RobertWilbrandt (approved):
Sorry about that, i was not aware that i needed to be the person to re-review this. Looks good now.

fmauch:
@Mergifyio backport iron

PR #733, opened 2023-07-06
Start ursim from lib

fmauch:
This is basically to test https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/155

Keeping this draft until the client_lib is merged so we can remove the rosinstall entry.


RobertWilbrandt (changes requested):
Seems like a great move to no longer have this redundancy, but i think we should adjust our documentation and tests along with the deprecation.

fmauch:
@RobertWilbrandt I've updated docs and tests accordingly.

PR #739, opened 2023-07-11
Use mock_hardware and mock_sensor_commands instead of fake

fmauch:
This has been deprecated a while back and was never adapted. This PR goes alongside https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/77


RobertWilbrandt (changes requested):
The documentation and the `ur_calibration` README still mention `fake_hardware`. Those places should get adjusted along with the launch files.

fmauch:
This should not necessarily get backported, as on humble this only a warning.

RobertWilbrandt (approved):

PR #749, opened 2023-07-20
Port execution tests to launch_pytest

RobertWilbrandt:
Our tests currently have some redundant parts like launch descriptions, service interfaces and initialization code. Before working on more tests i think we should clean that up and make our current tests more maintainable in the process. This should result in the following tasks:
- [x] Move launch descriptions for just the dashboard client and the full robot driver to `robot_launch_descriptions` so they can be resued among all tests.
- [x] Move core rclpy (init, node) into fixtures to replace `unittest` initialization structure.
- [ ] Move the common robot interfaces (to dashboard client, controller manager, common controllers, ...) into fixtures.
- [ ] Port all tests to use these fixtures.

This also enables easily adding more paremetrization options (like testing all robot models for future moveit tests).

I decided to use `logging` for all test output instead of (as previously) using the node logger. This allows us to only show logging output from our tests or from the tested ROS nodes based on pytest arguments.

I am not quite happy with having to include `*` from `robot_launch_descriptions` in every test file, but i could not find a way to use these launch fixtures without this that did not result in `ScopeError`s. The tests also always show a warning `Warning: There is no current event loop` at the end for me that i don't quite understand yet, but the tests still seem to work fine for me.

### DIFF
--- a/ur_robot_driver/test/robot_launch_descriptions.py
+++ b/ur_robot_driver/test/robot_launch_descriptions.py
@@ -33,6 +33,8 @@ from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
 from launch_pytest.actions import ReadyToTest
 from launch_ros.substitutions import FindPackagePrefix, FindPackageShare
 
+CONTROLLER_SPAWNER_TIMEOUT = 120
+
 
 @launch_pytest.fixture(scope="class")
 def launch_dashboard_client():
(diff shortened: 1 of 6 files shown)


ur_robot_driver/test/robot_launch_descriptions.py
```diff
@@ -0,0 +1,153 @@
...
+    class.
+    """
+    declared_arguments = []
+
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "ur_type",
+            default_value="ur5e",
+            description="Type/series of used UR robot.",
+            choices=["ur3", "ur3e", "ur5", "ur5e", "ur10", "ur10e", "ur16e"],
```
urrsk, line 55: suggest adding ur20 as well

RobertWilbrandt:
While adding more fixtures i continuously had problems with execution ordering. In particular, the ordering between new launch configurations and the rclpy fixtures seemed to be broken and i was not able to get the tests running reliably. Some pytest arguments like `--setup-plan` also just produced errors within launch_pytest.

Given that, i'll close this in favor of #849.

PR #753, opened 2023-07-21
Branch out for Iron

fmauch:
Since #739 effectively is a break in the user's API we should not forward the iron branch to this change. Hence, we need to branch out and add a mergify rule to make our life easier.

I left the iron workflows running in main PRs on purpose, as in fact iron and rolling are currently still fully compatible.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -7,21 +7,18 @@ pull_request_rules:
       backport:
         branches:
           - humble
+      label:
+        add:
+          - humble
 
-  - name: Backport to galactic branch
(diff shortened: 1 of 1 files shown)


fmauch:
@RobertWilbrandt It would be nice if you could have a look at this.

RobertWilbrandt (approved):
Sorry for the delay, looks good to me. I don't agree with the reasoning on the CI workflows (we test rolling anyway and this just seems like a weird inconsistency without any gain), but this PR doesn't really affect that.

PR #754, opened 2023-07-21
Cleanup branched out iron branch

fmauch:
Now that the iron branch lives on its own it is time to remove non-iron specifics from the branch.


RobertWilbrandt (approved):
Looks good and pretty complete. While looking at this i was wondering if there is any reason to still have `.github/dockerursim`, but that doesn't have to be part of this PR.

PR #755, opened 2023-07-21
Fix branches of upstream workspaces for semi-binary builds

fmauch:
Some branches were not correct anymore and removed description from binary build's upstream ws.


fmauch:
@RobertWilbrandt could this get merged?

PR #785, opened 2023-08-28
Added arguments to test launch

urmahp:
Added arguments when launching the test:
 - Added robot ip
 - Added launch_ursim to choose whether or not to launch ursim when running the test.
 - If the arguments are not provided the behavior is the same as before adding the arguments.

Ensure that the scaled joint trajectory controller is running, before sending the motion.


ur_robot_driver/test/robot_driver.py
```diff
@@ -126,22 +148,48 @@ def generate_test_description(tf_prefix):
...
+            "ur_type": ur_type,
+            "launch_rviz": "false",
+            "controller_spawner_timeout": str(TIMEOUT_WAIT_SERVICE_INITIAL),
+            "initial_joint_controller": "scaled_joint_trajectory_controller",
+            "headless_mode": "true",
+            "launch_dashboard_client": "false",
+            "start_joint_controller": "false",
+            "tf_prefix": tf_prefix,
+        }.items(),
+        condition=UnlessCondition(launch_ursim),
```
RobertWilbrandt, line 187: I don't particularly like having the robot driver include duplicated:
- Can you define the `robot_driver` once and then use that through a `launch.actions.GroupAction` for the non-ursim case and through `driver_starter` otherwise? That seems like a more compact way of achieving the same thing.
- More generally: Is there a reason for not wanting to wait for the dashboard server when not using ursim?

Either way, i would make sure that the driver behaves the same for both cases (in particular, that e.g. `robot_ip` is passed to both invocations).

urmahp: I will have a look at your suggestion to make it more compact, which I agree is a better idea. 

The reason for not waiting for the dashboard server and not using the IP as argument when launching the ursim is that the ip is hardcoded in the wait for dashboard server script and in the launch ursim script. But I could change it such that the configured ip is parsed to the scripts as well? However, this would require changes to the client library as well as the start usrsim scrip is located in the client library.

PR #804, opened 2023-09-18
Updated to ros2_controllers JTC API

fmauch:
This should fix the current build errors on Iron and Rolling semi binary.

I haven't verified that this is actually the working solution, hence the draft status.

The changes came in in https://github.com/ros-controls/ros2_controllers/pull/609


fmauch:
I gave this a manual test and the integration tests seem to be fine, as well, so making this ready for review.

urmahp (approved):
LGTM, I have also verified manually that it works.

fmauch:
I'm closing this in favor of #810 which actually is the better approach, as we are updating the complete API there.

PR #807, opened 2023-09-19
CI: Add labels from backports to target PR, not the original PR

fmauch:
I've configured that wrong in the past. Looking at the [docs](https://docs.mergify.com/actions/backport/) this should be the correct settings.


urmahp (approved):
LGTM

PR #811, opened 2023-09-20
Added a test that sjtc correctly aborts on violation of constraints

fmauch:
This is the Humble backport of #810. Since the JTC api changed on rolling recently, a direct backport isn't really applicable.


fmauch:
Tests succeed -> ready for review. Please note: I had to raise the limit of allowed deviation, as the implementation on Humble seems to not stop as fast as the one from Rolling. Do you think we should look into this in this PR or move that to a follow-up PR?

urmahp (approved):
Looks good to me. As these changes are much better than what we currently have I vote for merging this and then addressing the fact that it is not stopping as fast in a separate PR.

PR #837, opened 2023-10-10
Do not start urscipt_interface when using mock hardware

fmauch:
As the urscript_interface tries to establish a connection to a robot we should not attempt to start it when use_mock_hardware is set to true.

Since the startup will not fail but simply block forever, I did not add a failing test beforehand.


RobertWilbrandt (approved):
Looks good to me.

I am not quite sure what causes the failing test, it seems unrelated but we should try to follow up on that.

PR #849, opened 2023-10-17
Simplify tests

RobertWilbrandt:
Our runtime tests have accumulated quite a bit of duplicated code for starting the driver and interfacing with ROS interfaces. Before adding more tests (MoveIt, mock_hardware, ...) this should get refactored and cleaned up.

### DIFF
--- a/ur_robot_driver/test/robot_driver.py
+++ b/ur_robot_driver/test/robot_driver.py
@@ -109,10 +109,6 @@ class RobotDriverTest(unittest.TestCase):
     #
 
     def test_start_scaled_jtc_controller(self):
-        req = SwitchController.Request()
-        req.strictness = SwitchController.Request.BEST_EFFORT
-        req.activate_controllers = ["scaled_joint_trajectory_controller"]
-
         self.assertTrue(
             self._controller_manager_interface.switch_controller(
                 strictness=SwitchController.Request.BEST_EFFORT,
--- a/ur_robot_driver/test/robot_driver.py
+++ b/ur_robot_driver/test/robot_driver.py
@@ -28,6 +28,8 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 
+import os
+import sys
 import time
 import unittest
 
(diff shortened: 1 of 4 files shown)


fmauch:
Maybe this would be a good point to add the URSim log to the artifacts similar to https://github.com/UniversalRobots/Universal_Robots_Client_Library/commit/9b695e34a19dda897aa7e2fa5336d637d924567d

RobertWilbrandt:
That is not transferrable 100% - Because we start ursim from the test launch description the container is already dead when we want to copy out the logs. That is certainly solvable, but somewhat unrelated to restructuring the code. I'd prefer merging this now and adding that feature later on.

ur_robot_driver/test/robot_driver.py
```diff
@@ -223,9 +112,13 @@ def test_start_scaled_jtc_controller(self):
         req = SwitchController.Request()
         req.strictness = SwitchController.Request.BEST_EFFORT
         req.activate_controllers = ["scaled_joint_trajectory_controller"]
-        result = self.call_service("/controller_manager/switch_controller", req)
 
-        self.assertEqual(result.ok, True)
+        self.assertTrue(
+            self._controller_manager_interface.switch_controller(
+                strictness=SwitchController.Request.BEST_EFFORT,
+                activate_controllers=["scaled_joint_trajectory_controller"],
```
fmauch, line 119: This seems to be a duplication of the two lines above the assert statement. Probably those above should get deleted?

ur_robot_driver/test/robot_driver.py
```diff
@@ -246,13 +139,8 @@ def io_msg_cb(msg):
...
         test_pin = 0
 
-        set_io_req = SetIO.Request()
-        set_io_req.fun = 1
-        set_io_req.pin = test_pin
-        set_io_req.state = 1.0
-
-        self.node.get_logger().info(f"Setting pin {test_pin} to {set_io_req.state}")
-        self.call_service("/io_and_status_controller/set_io", set_io_req)
+        logging.info("Setting pin %d to 1.0", test_pin)
```
fmauch, line 142: Is there a reason you didn't use the `f"..."` notation? From my understanding that is "the recommended" way to format strings in Python nowadays.

fmauch (changes requested):
This does look mostly fine except for the code duplication mentioned.

This stands in conflict with #785. But I suggest, we first get this refactoring merged and then update #785 accordingly.

ur_robot_driver/test/robot_driver.py
RobertWilbrandt: My understanding was that logging specifically still recommends the '%' syntax as it can avoid transforming the arguments to strings if the log output is not actually used. Pylint has warnings for this ([logging-fstring-interpolation](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html)) which is why i stuck to this syntax for now, but i am open to changing this.

ur_robot_driver/test/robot_driver.py
RobertWilbrandt: Yes, that was an oversight on my part. Should be fixed now.

fmauch (approved):
Thank you @RobertWilbrandt This will make future test extensions much easier.

PR #850, opened 2023-10-18
Add test reports

fmauch:
Add a test report stage to the ICI. This should get a summary of our tests to the status page and show failed tests in PRs directly. Also, with this we will get statistics about flaky tests (if there are).


fmauch:
This still requires some work regarding making posts into PRs and job error output possible... Will postpone this for now.

PR #852, opened 2023-10-19
Remove noisy controller log message

RobertWilbrandt:
It does not provide a lot of value, but is printed every time the controllers are queried.

This even happens when calling `ros2 control list_controllers` or continuously while using `rqt_joint_trajectory_controller`.


fmauch (approved):
fine with me.

Issue #571, opened 2022-12-13
Publish tool0_controller frame


fmauch:
Currently, we don't publish the tcp pose that is reported from the robot. In ROS1 we used to publish this as `tool0_controller` which proved to be helpful either for debugging purposes or to publish the wrench on. I think it would make sense to add this also to the ROS2 driver.

urfeex:
This has actually been implemented in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1108

PR #856, opened 2023-10-20
Publish controller flange frame

fmauch:
This PR adds publishing the controller's flange frame as suggested in #571.

It still needs some cleanup work as the code is currently far from being optimal in terms of memory allocation. Also, I'd like to add some tests and probably also want to move the pose_broadcaster directly to ros2_controllers.


fmauch:
Closing this in favor of #1108

PR #871, opened 2023-11-08
ScaledJTC - Added support for setting speedscaling factor via topic

firesurfer:
Hi,
this PR adds support for setting the speed scaling factor via `~/speed_scaling_factor`.

The behavior can be activated by specifying `use_speed_scaling_topic_instead` in the controller configuration. Not specifying or setting the variable to false will keep the default behavior. 

This feature allows using the controller in multi arm setups or for usage with different hardware which doesn't have any `speed_scaling state`. 

EDIT: Just saw that I based the PR on the iron branch. If I should rebase it on the main branch let me know

EDIT2: As an alternative we could expose it as a parameter which would make a bit more similar to the `cartesian controllers`

EDIT3: The topic can now be changed via the config file. This allows to have one speed_scaling_factor topic for multiple controllers. Default is still `~/speed_scaling_factor`. 
The topic is now transient_local.

### DIFF
--- a/ur_controllers/include/ur_controllers/scaled_joint_trajectory_controller.hpp
+++ b/ur_controllers/include/ur_controllers/scaled_joint_trajectory_controller.hpp
@@ -45,11 +45,15 @@
 #include "rclcpp/duration.hpp"
 #include "scaled_joint_trajectory_controller_parameters.hpp"
 
+#include <std_msgs/msg/float64.hpp>
+
 namespace ur_controllers
 {
 class ScaledJointTrajectoryController : public joint_trajectory_controller::JointTrajectoryController
 {
 public:
+  using ScalingFactorMsg =  std_msgs::msg::Float64;
+
   ScaledJointTrajectoryController() = default;
   ~ScaledJointTrajectoryController() override = default;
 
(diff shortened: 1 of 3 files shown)


firesurfer:
@fmauch I might have also encountered a bug with the ScaledJTC but I am not sure if it comes from my adaptions. Therefore I add it here.

While testing we quite often ran into the `Aborted due goal_time_tolerance exceeding by 31.005506 seconds` error.
The motion is still performed as desired. 


```
```

firesurfer:
Okey I managed to fix the issue. It is unrelated to the actual change introduced in this PR. (I can separate if necessary into an extra PR). 

The Goal Time Tolerance Violation check if only executed if the arm is not within the position tolerances after the trajectory was finished sampling. So this issue will only occur if one runs fast movements with load (I dont know why I had it also in sim...). During such movements we won't be within the goal position tolerances and therefore we start checking the goal time tolerance. But the goal time tolerance did not compensate for the fact that if we scale down the trajectory execution that...well execution takes longer than expected. I hopefully fixed this

firesurfer:
See #883 for the same PR but based on main. For the issue described in this PR see #882

PR #872, opened 2023-11-09
Add backward_ros to driver

fmauch:
This should help debugging tremendously.


RobertWilbrandt (approved):
Agree 100%, great addition

PR #875, opened 2023-11-14
Use iron branches for ros2_control

fmauch:
same as #874


RobertWilbrandt (approved):
Great, it's good to stay up-to-date with these.

PR #883, opened 2023-11-16
Added option and implementation for setting speedscaling factor via topic

firesurfer:
Same as #871 but based on main

### DIFF
--- a/ur_controllers/include/ur_controllers/scaled_joint_trajectory_controller.hpp
+++ b/ur_controllers/include/ur_controllers/scaled_joint_trajectory_controller.hpp
@@ -45,11 +45,15 @@
 #include "rclcpp/duration.hpp"
 #include "scaled_joint_trajectory_controller_parameters.hpp"
 
+#include <std_msgs/msg/float64.hpp>
+
 namespace ur_controllers
 {
 class ScaledJointTrajectoryController : public joint_trajectory_controller::JointTrajectoryController
 {
 public:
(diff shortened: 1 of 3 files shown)


firesurfer:
@RobertWilbrandt might want to review this patch ?

firesurfer:
@fmauch  and @RobertWilbrandt  ?

firesurfer:
@fmauch  and @RobertWilbrandt  can you please take a look at the PR (and the other one related to it). I start running into merge conflicts now. I think the changes are not too large and are now tested on our system for roughly 2 months.

firesurfer:
@fmauch given that #882 is now merged you might want to take a look at this one ;)

firesurfer:
Could you elaborate why you decided to use a service instead of a latched topic. I think a latched has various advantageous such as 

* Consistency - when you load a controller later on you will get the currently set `speed scaling factor` 
* Any other node can directly obtain the speed scaling factor by subscribing to the topic at any point of time

Especially the first point is very important for our application where we have around 3-5 (scaled)JTCs running at the same time. If I were to you a service I would have to make sure to set the current scaling factor to each controller.

From my point of view it would make more sense to suggest using a latched topic upstream..

fmauch:
Thank you for the elaboration, we will discuss this upstream. Unfortunately I don't have an open PR to discuss this atm, as I still want to bring it to a more shaped state, first. But I'll add it to my list :-)

PR #896, opened 2023-12-19
Update JTC API

fmauch:
This should fix the semi-binary builds (and future testing builds) on Humble.


fmauch:
@RobertWilbrandt @VinDp merging this rather soon would be appreciated, since currently the ur_controller build on the buildfarm is failing. I know it's Christmas, but having this available afterwards rightaway would be nice.

VinDp (approved):
Checked and tested, everything looks fine to me.

PR #906, opened 2024-01-11
fix move_group_node crash during initialization

CalaW:
fix #903

Moveit2 recently refactored planning pipeline, we need to change the config file correspondingly. See https://github.com/ros-planning/moveit2/blob/1fb4041e690c64b847d2df1e844c8de148796c86/MIGRATION.md

- Change planning plugin to a list of plugins, according to the change introduced in 10/2023 in moveit rolling.
- Split request_adapters to request_adapters and response_adapters, according to the refactor of planning pipeline introduced in 10/2023 in moveit rolling.

I tested it locally with URSim. Planning and execution works well.

PR #913, opened 2024-01-25
Add Moveit startup tests

fmauch:
This should help us finding Moveit related errors like the one fixed in #906 earlier.

## ToDo:

- [x] Add basic startup test
- [ ] Fix servo_node startup (the test should fail for now, since that is actually broken)
- [ ] Add MoveIt! to semi-binary builds


fmauch:
I could not get the servo node running correctly, which is why I would suggest taking fixing this out of this PR. Maybe we don't even put [7619306](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/913/commits/76193064ebbc5e5ae4376bda8708d01fb9c4fdbf) in here?

CalaW:
I think we should rewrite the whole `ur_servo.yaml` according to https://github.com/ros-planning/moveit2/blob/d56a8b9464b1a6e026ceeb70485892c69fd5100a/moveit_ros/moveit_servo/config/servo_parameters.yaml. Seems they did a major refactor on moveit_servo and lots of servo parameters are deprecated

CalaW:
Also I tested servo on both URSim and a real UR5, with same changes in [7619306](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/913/commits/76193064ebbc5e5ae4376bda8708d01fb9c4fdbf). It worked well!

PR #916, opened 2024-01-29
Update `prefix` to `tf_prefix` for UR xacro

mrjogo:
Update the moveit launchfile in line with the UR xacro argument name change in https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/57.

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -114,7 +114,7 @@ def launch_setup(context, *args, **kwargs):
             " ",
             "output_recipe_filename:=rtde_output_recipe.txt",
             " ",
-            "prefix:=",
+            "tf_prefix:=",
             prefix,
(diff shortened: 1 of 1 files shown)


fmauch:
First of all, thank you for the input! I am currently in the progress of finishing a tutorial about integrating a UR into your own model in https://github.com/UniversalRobots/Universal_Robots_ROS_Tutorials/pull/12 and I am making some changes to this repo on the way in https://github.com/fmauch/Universal_Robots_ROS2_Driver/tree/playground_rsp_launch. This change will definitively be important in that context but I might like to change it even further, since we are re-organizing the launch-interface for Jazzy anyway. I'll come back to this.

fmauch:
The prefix was removed from the moveit setup in #998. Thus, this PR becomes obsolete. We re-use the description provided by the robot_state_publisher and don't load an individual description into the move_group anymore.

PR #920, opened 2024-01-31
[URDF] Fix initial value of speed scaling factor syntax

fmauch:
This should make using mock_hw working again on rolling. I didn't check for now whether this is required for humble and iron, as well.


ur_robot_driver/urdf/ur.ros2_control.xacro
```diff
@@ -84,8 +84,9 @@
 
       <!-- NOTE The following are joints used only for testing with mock hardware and will change in the future -->
       <gpio name="${tf_prefix}speed_scaling">
-        <state_interface name="speed_scaling_factor"/>
-        <param name="initial_speed_scaling_factor">1</param>
+        <state_interface name="speed_scaling_factor">
+          <param name="initial_value">1.0</param>
+          </state_interface>
```
VinDp, line 89:
```suggestion
        <state_interface name="speed_scaling_factor">
          <param name="initial_value">1.0</param>
        </state_interface>
```
Small detail: it would be better to remove the additional space to align the 2 lines.

fmauch:
Thanks for noticing @VinDp

VinDp (approved):

PR #933, opened 2024-02-15
Port robot_state_helper to ROS2

fdurchdewald:
This pull request includes the 'robot_state_helper' node, which has been ported from the ROS1 to the ROS2 version of the driver.


urfeex:
@fdurchdewald I hope you don't mind me coming back to this. As far as I remember you told me that there were still some unresolved problems with this, but I just rebased and tested this and it seems to to exactly what it is supposed to do.

URJala:
Tested on a real robot, and switching between Running and Idle seems to work well, but there are couple of things: 

1. Switching to Power Off from any other state causes the helper node to abort the goal with the message:
`
Robot reached higher mode than requested during recovery. This either means that something went wrong or that a higher mode was requested from somewhere else (e.g. the teach pendant.)
`
 and then crash with the message:
`
goal_handle attempted invalid transition from state ABORTED with event SUCCEED
` 
requiring a restart of the driver to use it again.

2. If you try to use it while not in remote control mode, it will print an explanation that you should switch to remote control and reconnect, but it does not abort the action that requested the state change, so it just hangs forever.

3. When changing from running to idle, it seems to completely restart the robot, going through many state changes. But that might just be the way to put the robot in idle, I am not sure. 


Otherwise it seems to work well with resuming and stopping programs.
A couple of other notes, that might be more review-related:
It might be nice to define the available modes as constants in the action definition, and I am not quite sure I understand the need for both a `stop_program` and a `play_program` option, when a program cannot be played in anything other than running anyway, and therefore neither option has any effect in other modes than running (as far as i can tell).

urfeex:
@URJala Thanks for testing this. I've refactored the code to address some of your findings.

> 1. Switching to Power Off from any other state causes the helper node to abort the goal with the message

That should be gone now.

> 2. If you try to use it while not in remote control mode, it will print an explanation that you should switch to remote control and reconnect, but it does not abort the action that requested the state change, so it just hangs forever.

I assume this is gone, as well, but I cannot test it currently.

> 3. When changing from running to idle, it seems to completely restart the robot, going through many state changes. But that might just be the way to put the robot in idle, I am not sure.

To my knowledge, the dashboard server doesn't contain a "engage brakes" call. So, we have to go through power_off for that transition.

Regarding stop and play: Did you read what I wrote into the documentation?

URJala:
Tested it again, and it seems to work perfectly now, couldnt find any of the problems from yesterday.

urfeex (approved):

PR #939, opened 2024-02-29
Use ros2 control node from controller_manager and description topic

fmauch:
Since CM's control node also supports realtime scheduling and we use a non-blocking read anyway, there's no need for keeping around our own control node anymore.

Also, I switched from the description parameter to the description topic, as the parameter will soon be removed upstream


destogl (approved):
Very nice and very good statistics of the PR :smile:

VinDp (approved):
Looks and works good.

PR #944, opened 2024-03-11
Forward trajectory controller

### LINKED ISSUE #839: Passthrough controllers
### Feature summary

We should implement a trajectory forwarding controller  similar to https://github.com/UniversalRobots/Universal_Robots_ROS_passthrough_controllers

We already have multiple possibilities to forward trajectories to the robot controller instead of interpolating it on the ROS machine and sending commands to the robot in each control cycle. [Recently](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/151), the ur_client_library got added the possibility to do spline interpolation on the robot controller similar to the one done by ros2_control's JTC.

Interpolating on the robot has the advantage of a lower real-time requirement of the driver, since only the trajectory points are sent to the robot in the beginning. For this to be possible, we also [recently](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/178) added the possibility to make the receive timeout on the robot depending on the actual control mode.

### Related issues
* https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/151
* https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/178
* https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/543

### Tasks
*To complete this issue involves*
- [ ] Implement the controller
- [ ] Make documentation
- [ ] Make Unit test
- [ ] Make example
- [ ] Test on real hardware

URJala:
The controller should be able to forward a cartesian tracjectory to the robot for interpolation.
Creates errors on compilation due to multiple definition of functions, when inheriting from the joint_trajectory_controller. But looking at the passthrough controllers from ROS1 that may not be the right controller to use anyway?
All inputs are welcome.

closes #839


RobertWilbrandt:
I cannot reproduce your build errors, are you up-to-date with the packages listed in `Universal_Robots_ROS2_Driver.rolling.repos`?

As for the general idea, i don't think the forwarding controller should be derived from the JTC at all as it doesn't do any spline interpolation itself. You can just derive directly from `controller_interface::ControllerInterface`.

The main problem we will have is encoding the trajectory in a way that it reaches the hardware interface. In ROS 1, we could just define `TrajectoryInterface` and use it in the hardware and controller. This is not really possible in ROS 2, as you access `LoanedStateInterface`s from the controller that only give you access to one double value each. I know that some people are working on allowing more complex interface types, but am also not sure if this would help us (as we ideally want to use data that does not have a statically defined size). @fmauch do you know about the current state of these efforts?

URJala:
The controller is now functional, and you should be able to test it out yourself. A user can give the controller a trajectory to forward via the action server running on the controller. It is possible to define trajectories with just positions, or with both positions and velocities and/or accelerations. The trajectory is transferred to the hardware interface via the command interfaces, before the hardware interface then transfers it to the robot. 
Only one trajectory will be accepted at a time, if one is executing others will be denied. 
The controller monitors the execution time of the trajectory, and gives a warning if it takes too long, although it isn't counting correctly currently (It warns too early). This has no effect on the execution of the trajectory.
When the robot driver is launched this controller is not running, and needs to be loaded/activated manually, while the scaled joint trajectory controller is unloaded, as they cannot run at the same time. The example `load_switch_controller_example.py` does this. The robot needs to be running while this is done, so the controller can be loaded correctly. (This should be fixed, so it doesn't need to run, but i don't know how currently.)
The examples `example_movej.py` and `example_spline.py` send a trajectory with only positions and one with positions, velocities and accelerations, respectively.
Let me know what you think of the construction of the controller and the communication between the controller and the hardware interface.

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -0,0 +1,407 @@
...
+    RCLCPP_ERROR(get_node()->get_logger(), "Can't accept new trajectories. Controller is not running.");
+    return rclcpp_action::GoalResponse::REJECT;
+  }
+
+  if (state_interfaces_[StateInterfaces::CONTROLLER_RUNNING].get_value() != 1.0) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Trajectory rejected, controller not running in hardware interface.");
+    return rclcpp_action::GoalResponse::REJECT;
+  }
+
+  if (command_interfaces_[CommandInterfaces::PASSTHROUGH_TRAJECTORY_TRANSFER_STATE].get_value() != 0.0) {
```
fmauch, line 188: Accessing interfaces directly from the action server callback thread seems dangerous.

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -0,0 +1,407 @@
...
+    return rclcpp_action::GoalResponse::REJECT;
+  }
+
+  // Check dimensions of the trajectory
+  if (check_dimensions(goal) == 0) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Trajectory rejected, dimensions of trajectory are incorrect.");
+    return rclcpp_action::GoalResponse::REJECT;
+  } else {
+    command_interfaces_[CommandInterfaces::PASSTHROUGH_TRAJECTORY_DIMENSIONS].set_value(
+        static_cast<double>(check_dimensions(goal)));
```
fmauch, line 199: Shouldn't this be done in the accepted callback?

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+  command_interfaces_[CommandInterfaces::PASSTHROUGH_TRAJECTORY_TRANSFER_STATE].set_value(1.0);
+
+  /* Start the executing thread of the action server, this will run until the trajectory is finished or cancelled. */
+  std::thread{ std::bind(&PassthroughTrajectoryController::execute, this, std::placeholders::_1), goal_handle }
+      .detach();
+  return;
+}
+
+void PassthroughTrajectoryController::execute(
```
fmauch, line 297: This seems like these checks should be done in the update method, instead, right?

ur_controllers/include/ur_controllers/passthrough_trajectory_controller.hpp
```diff
@@ -148,6 +162,9 @@ class PassthroughTrajectoryController : public controller_interface::ControllerI
   std::vector<control_msgs::msg::JointTolerance> goal_tolerance_;
   rclcpp::Duration goal_time_tolerance_ = rclcpp::Duration::from_nanoseconds(0);
 
+  realtime_tools::RealtimeBoxBestEffort<AsyncInfo> info_from_realtime_;
+  realtime_tools::RealtimeBoxBestEffort<AsyncInfo> info_to_realtime_;
```
fmauch, line 166: I think I don't quite understand why you want to use two real-time buffers. Wouldn't we want to use one for both directions?

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -233,17 +241,23 @@ controller_interface::return_type PassthroughTrajectoryController::update(const
...
     if (active_trajectory_elapsed_time_ > (max_trajectory_time_ + goal_time_tolerance_.seconds()) &&
         trajectory_active_) {
       RCLCPP_WARN(get_node()->get_logger(), "Trajectory should be finished by now. You may want to cancel this goal, "
                                             "if it is not.");
       trajectory_active_ = false;
     }
   }
+  AsyncInfo info = { command_interfaces_[CommandInterfaces::PASSTHROUGH_TRAJECTORY_TRANSFER_STATE].get_value(),
+                     command_interfaces_[PASSTHROUGH_TRAJECTORY_ABORT].get_value(),
+                     state_interfaces_[StateInterfaces::CONTROLLER_RUNNING].get_value(), true };
```
fmauch, line 256: Why do we need to get this information from the hardware interface? The controller should know by itself whether it is running or not?

urrsk:
- [ ] Remember to make documentation
- [ ] Add it to the list of controller in the ur_controller/index.rst

Universal_Robots_ROS2_Driver.jazzy.repos
```diff
@@ -13,8 +13,8 @@ repositories:
     version: foxy-devel
   ros2_control:
     type: git
-    url: https://github.com/ros-controls/ros2_control.git
-    version: master
+    url: https://github.com/pal-robotics-forks/ros2_control.git
+    version: fix/controller_update_period
```
fmauch, line 17:
```suggestion
    url: https://github.com/ros-controls/ros2_control.git
    version: master
```

Universal_Robots_ROS2_Driver.rolling.repos
```diff
@@ -13,8 +13,8 @@ repositories:
     version: foxy-devel
   ros2_control:
     type: git
-    url: https://github.com/ros-controls/ros2_control.git
-    version: master
+    url: https://github.com/pal-robotics-forks/ros2_control.git
+    version: fix/controller_update_period
```
fmauch, line 17:
```suggestion
    url: https://github.com/ros-controls/ros2_control.git
    version: master
```

VinDp (approved):
Works nicely and looks good, I just have some minor suggestions.

ur_controllers/doc/index.rst
```diff
@@ -135,3 +135,125 @@ Advertised services
...
   sensor.
+
+.. _passthrough_trajectory_controller:
+
+ur_controlers/PassthroughTrajectoryController
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This controller uses a ``control_msgs/FollowJointTrajectory`` action but instead of interpolating
+the trajectory on the ROS pc it forwards the complete trajectory to the robot controller for
+interpolation and execution. This way, the realtime requirements for the control PC.
```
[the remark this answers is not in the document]
fmauch: Looks like I was interrupted :-D I'll add something, thanks!

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -0,0 +1,625 @@
...
+    if (interface_it != state_interfaces_.end()) {
+      if (interface_it->get_interface_name() == "position") {
+        joint_position_state_interface_.emplace_back(*interface_it);
+
+      } else if (interface_it->get_interface_name() == "velocity") {
+        joint_velocity_state_interface_.emplace_back(*interface_it);
+      }
+    } else if (interface_it->get_interface_name() == "acceleration") {
+      joint_acceleration_state_interface_.emplace_back(*interface_it);
+    }
```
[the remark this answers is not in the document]
urfeex: Good catch!

ur_controllers/src/passthrough_trajectory_controller.cpp
```diff
@@ -0,0 +1,625 @@
...
+        end_goal();
+        return controller_interface::return_type::OK;
+      }
+    }
+
+    active_joint_traj_ = active_goal->gh_->get_goal()->trajectory;
+
+    if (current_index_ == 0 && current_transfer_state == TRANSFER_STATE_IDLE) {
+      active_trajectory_elapsed_time_ = rclcpp::Duration(0, 0);
+      rclcpp::Duration::from_seconds(duration_to_double(active_joint_traj_.points.back().time_from_start));
```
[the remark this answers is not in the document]
fmauch: thanks for noticing! I'll remove that line.

fmauch (approved):
Approving so my previous change request gets reset.

PR #950, opened 2024-03-20
Switch to "use_fake_hardware" for UR20

VinDp:
Currently the UR20 is the only one using "use_mock_hardware" and since the description pkg for humble still has "use_fake_hardware", it doesn't work launching it with the mock_hardware option. 
Maybe it's better to have fake_hardware also for the UR20 and in the future switch all the models together when fake_hardware can't be used anymore.

Also, the same goes for iron, so if this makes sense I will open a PR for the iron branch as well.


fmauch (approved):
That definitively does make sense on Humble.

fmauch:
@Mergifyio backport iron

PR #953, opened 2024-03-24
do not start tool communication if using mock hardware

aatb-ch:
the tool communication node is started without taking consideration of the `use_mock_hardware` flag, this adds it to the condition for starting, otherwise you'd see that when starting mock hardware:

```
[ur_ros2_control_node-1] [INFO] [1711293300.071767372] [io_and_status_controller]: Waiting for system interface to initialize...
[ur_ros2_control_node-1] [INFO] [1711293300.121983638] [io_and_status_controller]: Waiting for system interface to initialize...
[ur_ros2_control_node-1] [INFO] [1711293300.172223320] [io_and_status_controller]: Waiting for system interface to initialize...
...
```

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -260,7 +260,7 @@ def launch_setup(context, *args, **kwargs):
 
     tool_communication_node = Node(
         package="ur_robot_driver",
-        condition=IfCondition(use_tool_communication),
+        condition=IfCondition(use_tool_communication) and UnlessCondition(use_mock_hardware),
         executable="tool_communication.py",
         name="ur_tool_comm",
         output="screen",


aatb-ch:
mh, seems like i got totally confused with the `use_mock_hardware` / `use_fake_hardware` change, the proposed fix seems unrelated.

On the other hand installing from apt `ros-rolling-ur`, seems like the launch files still only react to `use_fake_hardware`, using `use_mock_hardware` per instructions crash on launch, and MoveIt still tries to use the `scaled_joint_trajectory_controller`:

starting with `ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_mock_hardware:=true launch_rviz:=false initial_joint_controller:=joint_trajectory_controller` shows load of error and crash, while using `ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur5e robot_ip:=yyy.yyy.yyy.yyy use_fake_hardware:=true launch_rviz:=false initial_joint_controller:=joint_trajectory_controller` cleanly starts.

then using `ros2 launch ur_moveit_config ur_moveit.launch.py ur_type:=ur5e launch_rviz:=true use_mock_hardware:=true` shows:

`[move_group-1] [ERROR] [1711297212.954714289] [moveit_ros.trajectory_execution_manager]: Failed to send trajectory part 1 of 1 to controller scaled_joint_trajectory_controller` when the controller should have been replaced with `joint_trajectory_controller`

Is the apt release so outdated from sources? Seems like something is out of sync.

aatb-ch:
Ok figured it out, the machine had both `ros-humble-ur` and `ros-rolling-ur` installed..

aatb-ch:
Last thing I might add, the current readme reads quite authoritative as it states:

**Install the driver**
`sudo apt-get install ros-rolling-ur`

which i followed again after not working on this project for a couple of months, which lead me completely astray since i'm on humble and forgot i had already installed the driver. The detailed docs on the contrary state:

**Install the driver using**
`sudo apt-get install ros-${ROS_DISTRO}-ur`

which seems quite more sensible, shouldn't that be rather the default install procedure? Should i open a new PR with this change on the readme?

RobertWilbrandt:
The README should state the correct distribution for each branch - when you are e.g. on the `humble` branch, it says `sudo apt-get install ros-${ROS_DISTRO}-ur`. This seems to be incorrect on the `iron` branch though.

On `humble`, we do state `This branch supports only ROS2 Humble. For other ROS2 versions, please see the respective branches.` just before the driver installation. It might make sense to carry this over to the other versions. @VinDp Maybe we should take a general look at making the READMEs more consistent between versions again.

aatb-ch:
Agreed, the humble branch states it correctly, explicitly with ${ROS_DISTRO}. When landing on the main branch the install states though:

`sudo apt-get install ros-rolling-ur` though, which i blindly followed since it felt quite authoritative, i usually am used to see `sudo apt-get install ros-${ROS_DISTRO}-ur`, so I actually thought, "he, funny, maybe there's a reason to it and I should not install the humble package somehow", which lead me down that wrong path. I could only advice to use also ${ROS_DISTRO} on the main branch maybe?

PR #940, opened 2024-02-29
Added service to enable and disable tool contact

URJala:
Also added service to get robot software version.

PR #964, opened 2024-04-11
Service to get software version of robot

URJala:
This PR implements a service to get the software version of the robot (useful for certain functionalities, for example tool contact can only work on e-series robots). This PR depends on [PR](https://github.com/ros-industrial/ur_msgs/pull/25) for functionality and [PR](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/940) for the tests to work.


ur_robot_driver/urdf/ur.ros2_control.xacro
```diff
@@ -223,6 +223,15 @@
         <command_interface name="zero_ftsensor_async_success"/>
       </gpio>
 
+      <gpio name="${tf_prefix}get_version">
+        <command_interface name="get_version_cmd"/>
+        <command_interface name="get_version_async_success"/>
+        <command_interface name="get_version_major"/>
+        <command_interface name="get_version_minor"/>
+        <command_interface name="get_version_bugfix"/>
+        <command_interface name="get_version_build"/>
```
fmauch, line 232: These should probably be state interfaces instead of command interfaces.

ur_robot_driver/test/test_common.py
```diff
@@ -240,7 +240,12 @@ class IoStatusInterface(
     _ServiceInterface,
     namespace="/io_and_status_controller",
     initial_services={"set_io": SetIO},
-    services={"resend_robot_program": Trigger},
+    services={
+        "resend_robot_program": Trigger,
+        "start_tool_contact": Trigger,
+        "end_tool_contact": Trigger,
+        "get_version": GetVersion,
```
fmauch, line 247:
```suggestion
        "get_robot_software_version": GetVersion,
```

ur_robot_driver/test/robot_driver.py
```diff
@@ -249,6 +249,20 @@ def test_trajectory_scaled(self, tf_prefix):
...
+
+            end_result = self._io_status_controller_interface.end_tool_contact()
+            self.assertEqual(end_result.success, False)
+        else:
+            start_result = self._io_status_controller_interface.start_tool_contact()
+            self.assertEqual(start_result.success, True)
+
+            end_result = self._io_status_controller_interface.end_tool_contact()
+            self.assertEqual(end_result.success, True)
+
```
fmauch, line 265: This doesn't seem like it belongs here.

fmauch (commented):
I think we should come up with a better name than "GetVersion" in the message, the code and the interfaces. I proposed `get_robot_software_version` at some point, but maybe `get_polyscope_version` might also be good. What do you think @URJala @VinDp @urrsk ?

ur_robot_driver/test/robot_driver.py
URJala: This is a re-implementation of the tool contact test, so it will work on all versions of the robot. The one included in the [Tool contact PR](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/940) will only work on e-series robots, as it cannot see the robot's software version, and tool contact can only be used on e-series robots. 
I could move this to it's own pull request if that makes more sense? I agree that it probably doesn't quite fit in here.

URJala: @fmauch I have removed the tool contact parts from both pull requests now. I will make a new PR to add the tests back when this has been merged.

ur_controllers/include/ur_controllers/gpio_controller.hpp
```diff
@@ -79,6 +80,8 @@ enum CommandInterfaces
   ZERO_FTSENSOR_ASYNC_SUCCESS = 32,
   HAND_BACK_CONTROL_CMD = 33,
   HAND_BACK_CONTROL_ASYNC_SUCCESS = 34,
+  GET_VERSION_CMD = 35,
+  GET_VERSION_ASYNC_SUCCESS = 36
```
fmauch, line 84: Thinking about this... I don't think, we actually need command interfaces. The information that we want to publish, doesn't change over the runtime of this controller. We could initially fill the values inside the hardware interface when configuring the hardware and then just fill the service response from those.

ur_controllers/include/ur_controllers/gpio_controller.hpp
fmauch, line 1: Let's take this opportunity and move things to a new `ur_configuration_controller`. We've put a lot of things into this controller in the past that has nothing to do with IOs.

ur_robot_driver/doc/migration/jazzy.rst
fmauch, line 1: This looks like it got in here accitentally

ur_moveit_config/launch/ur_moveit.launch.py
fmauch, line 1: I think this got in here accidentally.

fmauch (changes requested):

ur_controllers/src/gpio_controller.cpp
fmauch, line 1: This should probably go.

ur_controllers/src/ur_configuration_controller.cpp
```diff
@@ -0,0 +1,122 @@
...
+//----------------------------------------------------------------------
+
+#include <ur_controllers/ur_configuration_controller.hpp>
+namespace ur_controllers
+{
+
+controller_interface::CallbackReturn URConfigurationController::on_init()
+{
+  param_listener_ = std::make_shared<ur_configuration_controller::ParamListener>(get_node());
+  params_ = param_listener_->get_params();
```
fmauch, line 48: It seems fair to move this to `on_configure()`

ur_controllers/src/ur_configuration_controller.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+controller_interface::InterfaceConfiguration URConfigurationController::command_interface_configuration() const
+{
+  // No command interfaces currently
+  controller_interface::InterfaceConfiguration config;
+  config.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+
+  const std::string tf_prefix = params_.tf_prefix;
+
```
fmauch, line 65:
```suggestion
```

ur_controllers/src/ur_configuration_controller.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+  return controller_interface::return_type::OK;
+}
+
+controller_interface::CallbackReturn
+URConfigurationController::on_activate(const rclcpp_lifecycle::State& /* previous_state */)
+{
+  get_robot_software_version_srv_ = get_node()->create_service<ur_msgs::srv::GetRobotSoftwareVersion>(
+      "~/get_robot_software_version", std::bind(&URConfigurationController::getRobotSoftwareVersion, this,
+                                                std::placeholders::_1, std::placeholders::_2));
```
fmauch, line 95: This should go to `on_configure()`. `on_activate()` is supposed to be real-time safe which creating a service definitively isn't.

ur_controllers/src/ur_configuration_controller.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  return controller_interface::CallbackReturn::SUCCESS;
+}
+
+controller_interface::CallbackReturn
+URConfigurationController::on_deactivate(const rclcpp_lifecycle::State& /* previous_state */)
+{
+  get_robot_software_version_srv_.reset();
```
fmauch, line 103: No need to do that here.

ur_controllers/include/ur_controllers/ur_configuration_controller.hpp
```diff
@@ -0,0 +1,89 @@
...
+namespace ur_controllers
+{
+
+enum StateInterfaces
+{
+  ROBOT_VERSION_MAJOR,
+  ROBOT_VERSION_MINOR,
+  ROBOT_VERSION_BUILD,
+  ROBOT_VERSION_BUGFIX
+};
```
fmauch, line 60: I think a `struct RobotSoftwareVersion` would be more fitting than an enum.

ur_controllers/src/ur_configuration_controller.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+bool URConfigurationController::getRobotSoftwareVersion(
+    ur_msgs::srv::GetRobotSoftwareVersion::Request::SharedPtr /*req*/,
+    ur_msgs::srv::GetRobotSoftwareVersion::Response::SharedPtr resp)
+{
+  resp->major = static_cast<uint32_t>(state_interfaces_[StateInterfaces::ROBOT_VERSION_MAJOR].get_value());
+  resp->minor = static_cast<uint32_t>(state_interfaces_[StateInterfaces::ROBOT_VERSION_MINOR].get_value());
+  resp->bugfix = static_cast<uint32_t>(state_interfaces_[StateInterfaces::ROBOT_VERSION_BUGFIX].get_value());
+  resp->build = static_cast<uint32_t>(state_interfaces_[StateInterfaces::ROBOT_VERSION_BUILD].get_value());
```
fmauch, line 114: This seems a bit dangerous. The `state_interfaces_` are usually filled in the realtime thread (inside the `update` function). You could put a data structure into a RealtimeBox and fill that in the `update` method (or maybe even the `activate` method would be sufficient?) and read from that box in a safe manner here.

ur_controllers/src/ur_configuration_controller_parameters.yaml
```diff
@@ -0,0 +1,6 @@
+ur_configuration_controller:
+   tf_prefix: {
+    type: string,
+    default_value: "",
+    description: "Urdf prefix of the corresponding arm"
```
fmauch, line 5:
```suggestion
    description: "URDF prefix of the corresponding arm"
```

ur_robot_driver/doc/migration/jazzy.rst
fmauch, line 1: This should not go into this PR I guess.

ur_robot_driver/test/robot_driver.py
```diff
@@ -92,6 +93,7 @@ def init_robot(self):
         self._dashboard_interface = DashboardInterface(self.node)
         self._controller_manager_interface = ControllerManagerInterface(self.node)
         self._io_status_controller_interface = IoStatusInterface(self.node)
+        self._configuration_controller_interface = ConfigurationInterface(self.node)
```
fmauch, line 96: As far as I can see this isn't used anywhere.

fmauch (changes requested):
This is definitively going in the right direction!

ur_controllers/include/ur_controllers/ur_configuration_controller.hpp
fmauch: This is actually not true.

ur_controllers/src/ur_configuration_controller.cpp
URJala: I chose to put it in the `on_activate` method. Used a RealtimeBox and a struct to hold the information.

ur_moveit_config/launch/ur_moveit.launch.py
fmauch, line 1: This file should not get removed by this PR. Please fix that.

ur_robot_driver/doc/migration/jazzy.rst
fmauch, line 1: This file should not get removed by this PR. Please fix that.

fmauch (changes requested):
I've changed the RealTimeBox with RealTimeBoxBestEffort. Apart from that, please see my other comments / requests.

ur_robot_driver/test/robot_driver.py
fmauch: Nevermind, I have been blind.

fmauch (approved):
This looks good to me. I'll merge this together with the message update.

PR #967, opened 2024-04-15
Update installation instructions for source build

christophfroehlich:
I think there is some information missing, if someone wants to build the driver from source on a non-rolling ROS distro.

Relates to https://robotics.stackexchange.com/questions/110644/ros2-ur-driver-package-failed-ur-controllers/110654#110654


fmauch (approved):
Thanks for pointing that out. I stumbled upon something similar in the ur_gz_simulation repo today.

PR #984, opened 2024-05-06
General docs

fmauch:
This PR updates documentation so it can be re-used by the assembled documentation that is currently being created.


ur_moveit_config/doc/index.rst
```diff
@@ -0,0 +1,27 @@
...
+================
+ur_moveit_config
+================
+
+This package contains an **example** MoveIt! configuration for Universal Robots arms. Since the
+default description contains only the arm this MoveIt! configuration package also only contains the
+arm without any objects around it.
+
+Since a robotic arm is usually not floating in empty space, but is mounted somewhere, it is also
+recommended to create a robot_description modelling the real-world scenario and to generate a
```
fmauch, line 12:
```suggestion
In a real-world scenario it is recommended to create a robot_description modelling the robot with its surroundings (Table where it is mounted on, objects in its environment, etc.) and to generate a
```

VinDp (approved):
Looks good, I have only some small suggestions.

PR #985, opened 2024-05-07
Remove dependency to docker.io

### LINKED ISSUE #769: Consider removing docker dependency from package.xml
I suggest removing 

```
  <test_depend>docker.io</test_depend>
```

As this is a very large dependency where many people prob. want to explicitly decide whether they install it or not. If implicitly installed it can even be a security risk on some systems especially if users are unaware of it.

There is `--skip-keys=SKIP_KEYS` for `rosdep`. But in order to specify this option one has to be aware of what is installed which might not be the case especially for quickly put together testing environments.

fmauch:
The dependency was basically added since this is needed to spawn up ursim. However, the start_ursim script catches the case that no docker executable is available and the tests requiring it are behind a compile definition which is off by default.

So, in order to not install docker for users automatically, the dependency shall be removed.

As a side-effect this has always stopped us from releasing the package for RHEL.

This closes #769

### DIFF
--- a/.github/workflows/reusable_ici.yml
+++ b/.github/workflows/reusable_ici.yml
@@ -51,3 +51,4 @@ jobs:
           ROS_DISTRO: ${{ inputs.ros_distro }}
           ROS_REPO: ${{ inputs.ros_repo }}
           CMAKE_ARGS: -DUR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=ON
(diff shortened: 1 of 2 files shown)


RobertWilbrandt (approved):
Looks good, as we don't even use docker for testing when `UR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS` is not set this does not make sense as a fixed dependency.

PR #990, opened 2024-05-11
Add servo node config to disable advertising /get_planning_scene

mrjogo:
Disables the servo node from advertising `/get_planning_scene`, which in the default launchfile, conflicts with `move_group`'s `/get_planning_scene`. See https://github.com/moveit/moveit2/issues/2731 for discussion.

Tested on a Humble branch.

### DIFF
--- a/ur_moveit_config/config/ur_servo.yaml
+++ b/ur_moveit_config/config/ur_servo.yaml
@@ -32,6 +32,12 @@ publish_joint_accelerations: false
 ## Plugins for smoothing outgoing commands
 smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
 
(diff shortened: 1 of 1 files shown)


fmauch:
Thank you @mrjogo that change seems reasonable. I'll check it and come back to this.

fmauch:
I just tested this and it seems to be fine. If you don't mind, I would include this commit into #998 for jazzy/rolling. This PR here would then be targeting iron and humble only.

mrjogo:
Sounds good. Let me know if you need me to do anything for that.

fmauch (approved):
Tested against iron and works fine.

fmauch:
@Mergifyio backport humble

Issue #917, opened 2024-01-30
Adding MoveItConfigsBuilder compatible example to ur_moveit_config


odbee:
### Feature summary

Hey, so it seems the current ur_moveit_config folder is not at all compatible with being used with  the MoveItConfigsBuilder. It would help greatly with running moveit directly through the C++ API if  one could take all the moveitconfigbuilder parameters and add them to the parameters of ones c++ API like this: `parameters=[moveit_config.to_dict()]`

### Tasks
*To complete this issue involves*
- [ ] Implement the feature
- [ ] Make documentation
- [ ] Make Unit test
- [ ] Make example
- [ ] Test on real hardware

VinDp:
Hey @odbee, 
thanks for the suggestion, we are currently working on that.
We are going to change the structure according to what is shown in this [tutorial page](https://github.com/UniversalRobots/Universal_Robots_ROS2_Tutorials/tree/main/my_robot_cell/my_robot_cell_moveit_config), do you have any opinions on that?

odbee:
Hey @VinDp thanks, for consering my Feature request, looks good! the structure looks good. I think it would be nice to have both scaled_joint_trajectory_controller and regular joint_trajectory_controller available in the "moveit_controllers.yaml", as well as different inverse kinematics  ( Trac-IK, Pic IK) and planning (pilz_industrial_motion_planner_planning_planner.yaml , ompl_planner.yaml) files,

fmauch:
Once issue I see with this is that the `robot_name` for the `robot_description` will be different for each robot model (e.g. `ur5e`. Because of that we currently use an xacro to generate the SRDF. As far as I understand the [MoveItConfigsBuilder](https://github.com/moveit/moveit2/blob/main/moveit_configs_utils/moveit_configs_utils/moveit_configs_builder.py) it doesn't support that but expects a finished srdf file.

PR #998, opened 2024-05-16
Make moveit_config compatible to moveit_configs_builder

fmauch:
As mentioned in #917 our current config is not compatible to the moveit_configs_builder.

 This PR aims at making our config compatible with it.

The current state is a working version which allows loading the description from the robot_state_publisher's topic and can be used in the moveit_configs_builder

## Things still need to be done
- [x] Lazy starting of move_group. Currently, it crashes when the description topic isn't yet published
- ~~[ ] Add upstream issue about the crash when no description topic is provided~~
- [x] Add moveit_servo

## Things that significantly change in this PR:
- the `tf_prefix` has been removed. It didn't work correctly, anyway. Since there are joints inside config files such as the `joint_limits.yaml` and they to my knowledge cannot handle argument substitution during loading inside the configs builder, this doesn't work.

## Things suggested in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/917#issuecomment-2100517142 I did not implement
- I did not add different IK solutions, as it seems not to be possible to pass them to the configs builder dynamically. As this is merely an example repo, I'd rather not add this. Users can still define their own when they implement their own launchfiles with the configs_builder in them.


gavanderhoorn:
Unorthodox perhaps, but what about not using the config builder at all and creating a more ROS-1-like MoveIt config pkg? Tyler made a nice example showing how to do this, right around the time I was also looking into it.

See [ROSCON 2023 - Retro ROS 2 Launch](https://tylerjw.dev/posts/xml-launch/).

fmauch:
Thank you for the feedback @gavanderhoorn!

We have considered for quite a while now whether we want to migrate our config package or not. Personally, I am also not the greatest fan of the way the configs builder hides the required config files away. Though once looking at its source code it became less hidden for me, so a better documentation could potentially improve things here. In my opinion there are a couple of arguments that justify making this step:

- configs-builder seems to be the standard used by MoveIt! nowadays. Of course, that doesn't mean that we'll have to follow that and we do a couple of things differently than a standard moveit_config package would, but it might help new users to get familiar with the structure more easily. And I expect that package to be used by less experienced MoveIt! users anyway, as more experienced users would probably create their own moveit_config package rightaway.
- Reusing this to write your own launchfiles launching your custom nodes should be much more convenient, especially if you are used to using the configs builder. The main point about this PR is changing the structure, not necessarily *using* the configs builder in our own launchfiles.  That's also (from my understanding) what the OP in #917 was expecting.
- Tyler's package I think is also from a time where the configs builder didn't exist, yet. The launchfiles become much less cluttered once the description is re-used from the description topic.
- Using one big config file for everything isn't necessarily reducing complexity in my opinion.

fmauch:
I decided against an upstream issue regarding the description topic as it seems intentional to have a [default of 10 seconds timeout](https://github.com/moveit/moveit2/blob/82ef3fc8dbf6770f309557924634ce0bc4ea6af8/moveit_ros/planning/rdf_loader/src/synchronized_string_parameter.cpp#L79).

ur_moveit_config/package.xml
```diff
@@ -6,31 +6,26 @@
...
 
-  <exec_depend>launch</exec_depend>
-  <exec_depend>launch_ros</exec_depend>
-  <exec_depend>moveit_kinematics</exec_depend>
-  <exec_depend>moveit_planners_ompl</exec_depend>
   <exec_depend>moveit_ros_move_group</exec_depend>
-  <exec_depend>moveit_ros_visualization</exec_depend>
-  <exec_depend>moveit_servo</exec_depend>
+  <exec_depend>moveit_kinematics</exec_depend>
+  <exec_depend>moveit_planners</exec_depend>
```
VinDp, line 22:
```suggestion
  <exec_depend>moveit_kinematics</exec_depend>
  <exec_depend>moveit_planners</exec_depend>
  <exec_depend>moveit_ros_visualization</exec_depend>
  <exec_depend>moveit_servo</exec_depend>
```

ur_moveit_config/package.xml
```diff
@@ -6,31 +6,26 @@
...
   <exec_depend>moveit_simple_controller_manager</exec_depend>
-  <exec_depend>rviz2</exec_depend>
+  <exec_depend>joint_state_publisher</exec_depend>
+  <exec_depend>moveit_configs_utils</exec_depend>
+  <exec_depend>moveit_ros_move_group</exec_depend>
   <exec_depend>ur_description</exec_depend>
-  <exec_depend>urdf</exec_depend>
-  <exec_depend>warehouse_ros_sqlite</exec_depend>
-  <exec_depend>xacro</exec_depend>
+
```
VinDp, line 28:
```suggestion
  <exec_depend>warehouse_ros_sqlite</exec_depend>
  <exec_depend>xacro</exec_depend>
  
```

ur_moveit_config/package.xml
```diff
@@ -6,31 +6,26 @@
...
   <maintainer email="wilbrandt@fzi.de">Robert Wilbrandt</maintainer>
 
   <license>Apache2.0</license>
 
+  <author email="denis@stoglrobotics.de">Denis Stogl</author>
   <author email="lovro.ivanov@gmail.com">Lovro Ivanov</author>
   <author email="zelenak@picknik.ai">Andy Zelenak</author>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
-  <buildtool_depend>ament_cmake_python</buildtool_depend>
```
VinDp, line 19:
```suggestion
  <buildtool_depend>ament_cmake_python</buildtool_depend>
  
  <exec_depend>launch</exec_depend>
  <exec_depend>launch_ros</exec_depend>
```

ur_moveit_config/CMakeLists.txt
```diff
@@ -1,15 +1,12 @@
...
 find_package(ament_cmake REQUIRED)
-find_package(ament_cmake_python REQUIRED)
 
-install(DIRECTORY config launch rviz srdf
+ament_package()
+
+install(
+  DIRECTORY launch
   DESTINATION share/${PROJECT_NAME}
 )
```
VinDp, line 11:
```suggestion
install(
  DIRECTORY config launch
  DESTINATION share/${PROJECT_NAME}
)
```

ur_moveit_config/CMakeLists.txt
```diff
@@ -1,15 +1,12 @@
...
-
-# Install Python modules
-ament_python_install_package(${PROJECT_NAME})
-ament_python_install_module(${PROJECT_NAME}/launch_common.py)
-
-ament_package()
+install(DIRECTORY config DESTINATION share/${PROJECT_NAME})
```
VinDp, line 12:
```suggestion
```

VinDp (approved):
Works nicely and looks good to me, I only have some small suggestions.

Issue #918, opened 2024-01-30
Passing "robot_calibration.yaml" not working via lauch-parameter "kinematics_config"


A13X15:
### Affected ROS2 Driver version(s)

Package: ros-humble-ur Version: 2.2.10-1jammy.20240126.081445

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with realtime patch

### How is the UR ROS2 Driver installed.

From binary packets

### Which robot platform is the driver connected to.

Real robot

### Robot SW / URSim version(s)

3.12.0

### How is the ROS driver used.

Through the robot teach pendant using External Control URCap

### Issue details

# Summary
Launching and passing the robot_calibration.yaml file via the launch-parameter kinematics_config does not work.

## Steps to Reproduce
The launch-command I used was
`ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur10 robot_ip:=my_robots_ip launch_rviz:=false initial_joint_controller:=joint_trajectory_controller kinematics_config:=/path/to/my_robot_calibration.yaml`

The calibration-file was extracted via the tutorial in this repo while the robot was turned on. I'm certain that my calibration file and path to it are correct.

When launching the driver, I get multiple deprecation warnings, one of them is about passing the robot description parameter. Once the ur_ros2_control_node-1 gets to the calibration checksum, it does not output the checksum of my yaml-file, but the default checksum 'calib_17227329492635474227' from the [default UR10 kinematics file](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/rolling/config/ur10/default_kinematics.yaml). So the correct calibration file is certainly not passed. This obviously results in an error (mismatched checksums).

## Expected Behaviour
I'd expect to be able to pass the calibration file via a simple launch parameter, similar to how it was on the ROS1 version of this driver. If that's not possible anymore, I'd like to know for certain what is the proper way of passing that calibration file.

# Workaround Suggestion
I'm aware of issue #711, I decided to create a new issue because it is not clear from that issue that passing the calibration file at launch does not work. I know that there are different ways of launching with a calibration file mentioned in #711 (modify the launch.py-file directly, or creating a separate description package based on the [Universal_Robots_ROS2_Description](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description)-Repo. Both of those options seem quite complicated and are not really well documented anywhere.

[...]

PR #1001, opened 2024-05-22
Add calibration file to launch arguments

VinDp:
At the moment passing the robot calibration as launch argument is not possible in Humble, but only on Rolling. These changes allow that and they should also fix [#918](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/918).


fmauch (approved):
Thanks @VinDp

Issue #1028, opened 2024-06-19
launching without a tf prefix is broken


fmauch:
### Affected ROS2 Driver version(s)

rolling master

### Used ROS distribution.

Rolling

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

Build both the ROS driver and UR Client Library from source

### Which robot platform is the driver connected to.

URSim in docker

### Robot SW / URSim version(s)

irrelevant

### How is the ROS driver used.

Through the robot teach pendant using External Control URCap

### Issue details

# Summary
Launching without the `tf_prefix` argument set doesn't work.

## Issue details
We removed that in 0e410cbabf7af13cf76713f78dbcb66e47058961 since it is read in the rsp.launch.py file now. However, it is not declared there.

## Steps to Reproduce
```
ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur5e robot_ip:=192.168.56.101
```

## Expected Behavior
It works

## Actual Behavior
```
[INFO] [launch]: All log files can be found below /home/mauch/.ros/log/2024-06-19-10-23-43-702967-ids-felino-2759927
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): launch configuration 'tf_prefix' does not exist
```

# Workaround Suggestion
- [ ] Add the launch argument to the `rsp.launch.py` file
- [ ] Create a test that is launching the driver without setting a `tf_prefix`

### Accept Public visibility

- [X] I agree to make this context public

PR #1029, opened 2024-06-19
Fix launching without a tf_prefix specified

fmauch:
This aims at fixing #1028, which is a regression introduced in #1020.

I first want the issue to show up in CI and then fix it with a consecutive commit.


VinDp (approved):
Looks good to me.

PR #1033, opened 2024-06-24
Also install srdf folder

fmauch:
Since we use the srdf from that folder, we should also install that.

This was broken in #998.


VinDp (approved):
Makes perfect sense, looks good to me.

Issue #1037, opened 2024-06-28
High CPU load on controller_stopper bringup and fails controller spawn.


kjjpc:
### Affected ROS2 Driver version(s)

Humble

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Docker

### How is the UR ROS2 Driver installed.

From binary packets

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

latest

### How is the ROS driver used.

Headless without using the teach pendant

### Issue details

# Summary
High CPU load on controller_stopper bring up.
Sometimes it causes the failure of controller's spawn.

## Issue details
Sometimes I encountered failures of spawn of controllers of UR.
I investigated the behavior and found the following failure process.

1. The controller_stopper calls list_controllers service on very high frequency. It causes high load on the controller manager of ROS2.
2. High load on the controller manager causes the service call failure from controller spawners.
3. The control spawners freeze waiting from the response from the controller manager.

There are 2 problems.
1. High frequency service call of controller_stopper of UR_ROS2_Driver.
2. Service call failure of ROS2 rcl on high load.

The high frequency service call is caused by while loop without sleep.
https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/humble/ur_robot_driver/src/controller_stopper.cpp#L76-L92

The problem of ROS2 rcl will be discussed on this issue.
https://github.com/ros2/rcl/issues/1163

## Steps to Reproduce
Bring up ur_control.launch.py
```
ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur5e robot_ip:=192.168.56.101
```

## Expected Behavior
Call the service every few tens of millseconds.

## Actual Behavior
Very high frequency service call occurs.

# Workaround Suggestion
Add a sleep in the loop.

### Accept Public visibility

- [X] I agree to make this context public

PR #1038, opened 2024-06-28
Add sleep between controller stopper's controller queries

fmauch:
Fixes the issue reported in #1037 

I've chosen a rather arbitrary 100ms of sleep. Maybe a shorter period would be desirable to avoid "long" periods where the controller seems to be running?

@kjjpc do you have an opinion on the sleep period?


kjjpc:
@fmauch
Thanks for fixing the issue.
100ms sounds good to me.

VinDp (approved):
Looks good to me.

Issue #980, opened 2024-04-24
Add Jazzy migration notes


fmauch:
We should add migration notes for Jazzy

- [x] Split description
- [x] rsp launchfile

PR #1045, opened 2024-07-02
Add migration notes for jazzy

fmauch:
This should be closing #980


VinDp (approved):
Looks very good to me.

PR #1049, opened 2024-07-08
Add force mode controller

URJala:
Added force mode arguments: damping and gain scaling to be usable with the force mode controller. damping and gain scaling will be changeable every time you start force mode when [PR](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/208) is implemented. 
This is an update to [PR](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/579)


URJala:
Also this is dependent on [PR](https://github.com/ros-industrial/ur_msgs/pull/26) for the service definitiion.

fmauch:
## Current ToDos:

- [ ] Claim all command interfaces
  Otherwise the robot might jump back once force mode is deactivated,  if there's another controller running.
- [ ] Make controller chainable so it can be combined e.g. with the JTC
- [ ] Add documentation
- [ ] Decide on the interface
  - Wrench / WrenchStamped?
  - Split limits vector into two? (Speed and deviation)

domire8:
Hello, exciting to see this work in progress, we from [AICA](https://www.aica.tech/) have already achieved force mode through ros2_control. My question is why you are implementing this through the async IOs and not as a full command mode in the `write` function? Is your goal to be able to have force mode active while another joint velocity controller is running, for instance?

Also, this PR is missing the updates in the xacro macros for the relevant command interface of the `force_mode` right?

Happy to have a discussion, if there is interest from your side

fmauch:
> Hello, exciting to see this work in progress, we from [AICA](https://www.aica.tech/) have already achieved force mode through ros2_control. My question is why you are implementing this through the async IOs and not as a full command mode in the `write` function? Is your goal to be able to have force mode active while another joint velocity controller is running, for instance?

Yes, exactly.

> Also, this PR is missing the updates in the xacro macros for the relevant command interface of the `force_mode`.

Yes, this is still active work. But thanks for the hint I'll add it to the list.

> Happy to have a discussion, if there is interest from your side

You don't happen to be at ROSCon next week?

domire8:
> I've been thinking about this comment and in all honesty, I don't think there is a valid reason why communication is managed outside of write. Especially, since we are doing the communication from the controller to the HW interface in the update loop, it seems that this is even a problem in terms of thread safety...

I can't say about thread safety, but I guess at the end the question is what you want to achieve. Our main goal for a force controller was to run force mode with an impedance controller as a primary control mode, while the robot has no other setpoints. For that, definitely the write function is the right place since the AsyncIO runs at a fixed (?) lower frequency. If however, you have applications in mind where one just enables force mode in one axis once and then uses joint velocity setpoints to drive the robot, the asyncio approach could work as well.

ur_controllers/src/force_mode_controller.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+  return controller_interface::return_type::OK;
+}
+
+bool ForceModeController::setForceMode(const ur_msgs::srv::SetForceMode::Request::SharedPtr req,
+                                       ur_msgs::srv::SetForceMode::Response::SharedPtr resp)
+{
+  // Reject if controller is not active
+  if (get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Can't accept new action goals. Controller is not running.");
```
VinDp, line 237:
```suggestion
    RCLCPP_ERROR(get_node()->get_logger(), "Can't accept new requests. Controller is not running.");
```

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -933,56 +997,96 @@ hardware_interface::return_type URPositionHardwareInterface::prepare_command_mod
...
+  if (std::any_of(start_modes_[0].begin(), start_modes_[0].end(),
                   [this](auto& item) { return (item == PASSTHROUGH_GPIO); }) &&
-      (std::any_of(start_modes_.begin(), start_modes_.end(),
+      (std::any_of(start_modes_[0].begin(), start_modes_[0].end(),
+                   [](auto& item) {
+                     return (item == hardware_interface::HW_IF_VELOCITY || item == hardware_interface::HW_IF_POSITION);
+                   }) ||
+       std::any_of(control_modes[0].begin(), control_modes[0].end(), [this](auto& item) {
+         return (item == hardware_interface::HW_IF_VELOCITY ||
+                 item == hardware_interface::HW_IF_POSITION /*|| item == PASSTHROUGH_GPIO*/);
```
VinDp, line 1037: Shouldn't this be uncommented?
```suggestion
                 item == hardware_interface::HW_IF_POSITION || item == PASSTHROUGH_GPIO);
```

ur_robot_driver/examples/force_mode.py
```diff
@@ -0,0 +1,136 @@
...
+if __name__ == "__main__":
+    rclpy.init()
+    node = Node("robot_driver_test")
+    robot = Robot(node)
+
+    # Activate force mode controller
+    robot.call_service(
+        "/controller_manager/switch_controller",
+        SwitchController.Request(
+            activate_controllers=["force_mode_controller", "scaled_joint_trajectory_controller"],
```
VinDp, line 62: Why are they both in `activate_controllers`?

VinDp (approved):
It looks good and works as intended, I just have few suggestions/questions :)

ur_robot_driver/examples/force_mode.py
urfeex: Fair point. I think I haven't overhauled the examples!

PR #1051, opened 2024-07-10
Add doc to custom URScript commands that it needs to be in headless mode

urrsk:
Seems to be an important detail missing that the custom URScript command only works in headless mode

### DIFF
--- a/ur_robot_driver/doc/usage.rst
+++ b/ur_robot_driver/doc/usage.rst
@@ -223,7 +223,7 @@ restarted again. Depending whether you use headless mode or not, you'll have to 
 external_control program again.
 
 .. note::
-  Custom URScript command only works in headless mode. And does not work with the external_control URCap.
+  On E-series robots or newer the robot needs to be in remote control mode in order to execute custom URScript commands.
   Currently, there is no feedback on the code's correctness. If the code sent to the
   robot is incorrect, it will silently not get executed. Make sure that you send valid URScript code!
 
--- a/ur_robot_driver/doc/usage.rst
+++ b/ur_robot_driver/doc/usage.rst
@@ -206,7 +206,7 @@ Custom URScript commands
 ------------------------
 
 The driver's package contains a ``urscript_interface`` node that allows sending URScript snippets
-directly to the robot when the robot is in headless/Remote control mode.
+directly to the robot when the robot is in remote control mode.
 
 It gets started in the driver's launchfiles by default. To use it, simply
 publish a message to its interface:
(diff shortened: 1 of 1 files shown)


ur_robot_driver/doc/usage.rst
```diff
@@ -221,6 +223,7 @@ restarted again. Depending whether you use headless mode or not, you'll have to
 external_control program again.
 
 .. note::
+  Custom URScript command only works in headless mode. And does not work with the external_control URCap.
```
fmauch, line 226: I think, this  is  a bit misleading, as in fact it is working together with the external_control URCap. The program just has to be restarted afterwards, as stated above.

ur_robot_driver/doc/usage.rst
```diff
@@ -206,7 +206,9 @@ Custom URScript commands
 ------------------------
 
 The driver's package contains a ``urscript_interface`` node that allows sending URScript snippets
-directly to the robot. It gets started in the driver's launchfiles by default. To use it, simply
+directly to the robot when the robot is in headless/Remote control mode.
```
fmauch, line 209: Wether or not we use headless mode doesn't really matter. But yes, we should state more clearly, that the robot will have to be in remote control mode for a PS5 robot.

```suggestion
directly to the robot when the robot is in remote control mode.
```

ur_robot_driver/doc/usage.rst
urrsk: @fmauch No that is not correct. On the e-series it only will work when the robot in remote control.
Maybe that is what we should write

ur_robot_driver/doc/usage.rst
urrsk: True it the remote control that makes it accepting code from external on Polyscope5

ur_robot_driver/doc/usage.rst
urrsk, line 226:
```suggestion
  Custom URScript command only works in remote control mode on E-series robots or later.
```

fmauch: How about: "On E-series robots or later the robot needs to be in remote control mode in order to execute custom URScript commands."

gavanderhoorn: > "On E-series robots or *newer* the robot needs to be in remote control mode in order to execute custom URScript commands."

urrsk:
```suggestion
  On E-series robots or newer the robot needs to be in remote control mode in order to execute custom URScript commands.
```

URJala (approved):

fmauch (approved):
Thank you @urrsk!

PR #1052, opened 2024-07-10
[doc] Update required polyscope version

fmauch:
this seems like it was never updated. We should think about linking this to the library's requirement section, but this is a quick fix for now.


VinDp (approved):
LGTM

PR #1055, opened 2024-07-10
[doc] Add more documentation regarding usage

fmauch:
This adds a note about installing `ros2controlcli` and an example move script.


urrsk:
Suggest running the example_move.py  in the CI

urrsk (approved):

fmauch:
> Suggest running the example_move.py in the CI or make an new issue for doing so and add some check in the example as well

Yes, let's do that!

PR #1064, opened 2024-07-18
Updated the UR family photo on the readme

urrsk:
Update the picture to also include the UR20 and UR30


fmauch (approved):
Thanks for providing this!

Issue #1062, opened 2024-07-17
Build from source is broken for Humble by latest updates in ros2_controllers package


cosmac:
### Affected ROS2 Driver version(s)

2.2.14

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

Build the driver from source and using the UR Client Library from binary

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

5.1

### How is the ROS driver used.

Headless without using the teach pendant

### Issue details

# Summary
Failed to build Humble branch from source. External function signature changed (see `get_segment_tolerances()` from [ros2-controllers](https://github.com/ros-controls/ros2_controllers/blob/humble/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp#L96)).

## Issue details
- Please see https://github.com/ros-controls/ros2_controllers/commit/112df35d9b1fd4353035a9bbac1c1a256b7eec15 for more details on recent changes to ros2-controllers

## Steps to Reproduce
```
git clone --branch humble https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver.git src/Universal_Robots_ROS2_Driver
vcs import src --skip-existing --input src/Universal_Robots_ROS2_Driver/Universal_Robots_ROS2_Driver.humble.repos

colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
```

## Expected Behavior
Build from source successful for all packages.

## Actual Behavior
Build fails during `ur_controllers`

# Workaround Suggestion
Update function call:
`default_tolerances_ = get_segment_tolerances(get_node()->get_logger(), params_);`

### Relevant log output

```shell
171.3 --- stderr: ur_controllers
171.3 /root/workspace/ros_ur_driver/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp: In member function 'virtual controller_interface::return_type ur_controllers::ScaledJointTrajectoryController::update(const rclcpp::Time&, const rclcpp::Duration&)':
171.3 /root/workspace/ros_ur_driver/src/Universal_Robots_ROS2_Driver/ur_controllers/src/scaled_joint_trajectory_controller.cpp:92:49: error: no matching function for call to 'get_segment_tolerances(joint_trajectory_controller::Params&)'
[...]

PR #1067, opened 2024-07-26
Updated scaled JTC to latest JTC updates

fmauch:
This should fix the current builds and #1062.


fmauch:
I'll merge this despite an approving review is missing since this is critical to resolve a regression and all possible reviewers are currently on vacation.

PR #1087, opened 2024-08-27
Update get_state to get_lifecycle_state

VinDp:
This PR updates the name of `get_state()` according to the [commit](https://github.com/ros-controls/ros2_controllers/commit/ce12694c6c948509a470783fd3b770cfab60a26e) that changes it to `get_lifecycle_state()` in ros2_controllers. In this way the CI for rolling and jazzy (Semi Binary) won't fail during building.


VinDp:
Since next sync is not soon, I think we can hold on this one.

fmauch (approved):
Very happy that we have this specific in ros2_control now :-)

PR #1088, opened 2024-08-28
Update maintainers team

VinDp:
Since I am going through the steps of doing a binary release of the [GZ simulation package](https://github.com/UniversalRobots/Universal_Robots_ROS2_GZ_Simulation), I need first of all to be added into the release team on `ros2-gbp` and, to create an issue for that, it's necessary to show that you belong to the maintainers team. 
So I updated things around also moving Robert to the authors section :)

### DIFF
--- a/ur/package.xml
+++ b/ur/package.xml
@@ -6,7 +6,6 @@
   <description>Metapackage for universal robots</description>
   <maintainer email="exner@fzi.de">Felix Exner</maintainer>
   <maintainer email="dipentima@fzi.de">Vincenzo Di Pentima</maintainer>
-  <maintainer email="wilbrandt@fzi.de">Robert Wilbrandt</maintainer>
   <license>BSD-3-Clause</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
--- a/ur/package.xml
+++ b/ur/package.xml
@@ -5,6 +5,7 @@
   <version>2.4.9</version>
   <description>Metapackage for universal robots</description>
   <maintainer email="exner@fzi.de">Felix Exner</maintainer>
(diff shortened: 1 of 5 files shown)


ur/package.xml
```diff
@@ -5,6 +5,7 @@
   <version>2.4.9</version>
   <description>Metapackage for universal robots</description>
   <maintainer email="exner@fzi.de">Felix Exner</maintainer>
+  <maintainer email="dipentima@fzi.de">Vincenzo Di Pentima</maintainer>
   <maintainer email="wilbrandt@fzi.de">Robert Wilbrandt</maintainer>
```
fmauch, line 9: Maybe it makes sense to remove Robert here in the same run. The package itself doesn't contain anything so it doesn't make sense to keep him as an author there, in my opinion.

ur_controllers/package.xml
```diff
@@ -7,13 +7,14 @@
 
   <maintainer email="denis@stoglrobotics.de">Denis Stogl</maintainer>
```
fmauch, line 8: Could we move @destogl to the authors section in that PR, as well?

ur_robot_driver/package.xml
```diff
@@ -7,13 +7,14 @@
 
   <maintainer email="denis@stoglrobotics.de">Denis Stogl</maintainer>
```
fmauch, line 8: Please also move @destogl to the authors section.

fmauch (changes requested):

ur/package.xml
VinDp: I was of the same opinion, but I was not sure it was the right thing to do. I will do so :+1:

destogl (approved):

PR #1089, opened 2024-08-28
Added option to publish SRDF file.

v-marsh:
Added argument to the launch file that can toggle the ability to publish the contents of the SRDF on the topic /robot_description_semantic. This allows the launch file to be used with custom moveit2 projects without having to source the SRDF for every launched node.

See following link for details: https://moveit.picknik.ai/humble/doc/examples/urdf_srdf/urdf_srdf_tutorial.html#string-topic-specification

Note: The default behaviour of not publishing this topic has been maintained, but now users have the option publish it if needed.

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -52,6 +52,7 @@ def launch_setup(context, *args, **kwargs):
     # General arguments
     description_package = LaunchConfiguration("description_package")
     description_file = LaunchConfiguration("description_file")
+    _publish_robot_description_semantic = LaunchConfiguration("publish_robot_description_semantic")
     moveit_config_package = LaunchConfiguration("moveit_config_package")
     moveit_joint_limits_file = LaunchConfiguration("moveit_joint_limits_file")
     moveit_config_file = LaunchConfiguration("moveit_config_file")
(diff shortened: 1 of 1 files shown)


fmauch (commented):
This seems to be similar to #1036. Is there a reason why you by default didn't enable this?

v-marsh:
Because #1036 only applies to the main branch which fails to build on ros2 humble. This request applies that same change to the humble branch which is useful for projects still running on humble due to its long term support.

fmauch:
Sorry, my comment was misleading. I was basically making two statements. The first being the backport of #1036. I'll make my second statement in-code to make it clear.

ur_moveit_config/launch/ur_moveit.launch.py
```diff
@@ -312,6 +316,13 @@ def generate_launch_description():
             description="URDF/XACRO description file with the robot.",
         )
     )
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "publish_robot_description_semantic",
+            default_value="False",
```
fmauch, line 322:
```suggestion
            default_value="True",
```

In #1036 (targeting rolling) we added this feature enabled by default. I'd suggest to make it the same here, unless you had a specific reason to default to `False`?

PR #1099, opened 2024-09-11
Prerelease/2.4.10

fmauch:
As usual, please rebase-merge, I'll push the tags and bloom-release later.

Now that ros2_control is being released (https://github.com/ros/rosdistro/pull/42800), we can also make a release to avoid having a broken buildfarm.


VinDp (approved):
Everything looks good.

PR #1102, opened 2024-09-16
Fix ci_status package reference for Jazzy

fmauch:
The status page for the Jazzy binary packages was missing referring to the Rolling packages instead of Jazzy.

Rbin -> Jbin


VinDp (approved):
Looks good to me.

Issue #1103, opened 2024-09-17
xacro files with comments are interpreted as YAML


katallen405:
### Affected ROS2 Driver version(s)

2.4.10

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with realtime patch

### How is the UR ROS2 Driver installed.

From binary packets

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

all

### How is the ROS driver used.

Through the robot teach pendant using External Control URCap

### Issue details

# Summary
xacro files can be interpret as xacro when using description_package

## Issue details
When launching with a custom description package (to show gripper), it causes the following error:
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Unable to parse the value of parameter robot_description as yaml. If the parameter is meant to be a string, try wrapping it in launch_ros.parameter_descriptions.ParameterValue(value, value_type=str)

This ROS issue has a known bug and workaround, but it requires that you change something in the file: 
https://answers.ros.org/question/417369/caught-exception-in-launch-see-debug-for-traceback-unable-to-parse-the-value-of-parameter-robot_description-as-yaml/ 

/ur_control.launch.py
https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/launch/ur_control.launch.py

## Steps to Reproduce
xacro files with commented xacro lines are particularly prone to this issue

## Expected Behavior
valid xacro should launch correctly

## Actual Behavior
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Unable to parse the value of parameter robot_description as yaml. If the parameter is meant to be a string, try wrapping it in launch_ros.parameter_descriptions.ParameterValue(value, value_type=str)

*What did you observe? If possible please attach relevant information.*

# Workaround Suggestion
no workaround

### Relevant log output

```shell
ros2 launch ur_robot_driver ur3e.launch.py robot_ip:=130.64.17.5 launch_rviz:=false description_package:=ceeorobot_description --debug
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.IncludeLaunchDescription'
[INFO] [launch]: All log files can be found below /home/kat/.ros/log/2024-09-17-11-30-42-642663-ur3e-hub-151123
[INFO] [launch]: Default logging verbosity is set to DEBUG
[...]

PR #1106, opened 2024-09-18
Assure the description is loaded as string

fmauch:
If this isn't explicitly specified, the description string might be interpreted as a yaml content, which leads to problems, obviously.

This was mentioned in #1103 and needs to get backported. However, since we refactored loading the description this will not be a straightforward backport, I'll open a separate PR.


VinDp (approved):
Looks good to me.

PR #1107, opened 2024-09-18
Assure the description is loaded as string

fmauch:
If this isn't explicitly specified, the description string might be interpreted as a yaml content, which leads to problems, obviously.

Implements #1103


VinDp (approved):
Looks good to me.

fmauch:
@Mergifyio backport humble

PR #1108, opened 2024-09-19
Use pose_broadcaster to publish the TCP pose

fmauch:
This uses the PoseBroadcaster currently written by @RobertWilbrandt. As he wants to contribute that to ros2_controllers, this will stay in a draft state until it is merged upstream.


Universal_Robots_ROS2_Driver.jazzy.repos
```diff
@@ -5,8 +5,8 @@ repositories:
     version: master
   Universal_Robots_ROS2_Description:
     type: git
-    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
-    version: rolling
+    url: https://github.com/fmauch/Universal_Robots_ROS2_Description.git
+    version: pose_broadcaster
```
fmauch, line 9:
```suggestion
    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
    version: rolling
```

Universal_Robots_ROS2_Driver.jazzy.repos
```diff
@@ -17,8 +17,8 @@ repositories:
     version: master
   ros2_controllers:
     type: git
-    url: https://github.com/ros-controls/ros2_controllers
-    version: master
+    url: https://github.com/RobertWilbrandt/ros2_controllers
+    version: pose_broadcaster
```
fmauch, line 21:
```suggestion
    url: https://github.com/ros-controls/ros2_controllers
    version: master
```

Universal_Robots_ROS2_Driver.rolling.repos
```diff
@@ -5,8 +5,8 @@ repositories:
     version: master
   Universal_Robots_ROS2_Description:
     type: git
-    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
-    version: rolling
+    url: https://github.com/fmauch/Universal_Robots_ROS2_Description.git
+    version: pose_broadcaster
```
fmauch, line 9:
```suggestion
    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
    version: rolling
```

Universal_Robots_ROS2_Driver.rolling.repos
```diff
@@ -17,8 +17,8 @@ repositories:
     version: master
   ros2_controllers:
     type: git
-    url: https://github.com/ros-controls/ros2_controllers
-    version: master
+    url: https://github.com/RobertWilbrandt/ros2_controllers
+    version: pose_broadcaster
```
fmauch, line 21:
```suggestion
    url: https://github.com/ros-controls/ros2_controllers
    version: master
```

gavanderhoorn:
High-level question: would it perhaps be better to have the hardware interface offer a quaternion instead of an Euler angle triplet?

That would seem cleaner to me, in the sense of it being more of an `N-to-1` mathematical abstraction.

While they aren't necessarily very human friendly, quaternions have definite benefits mathematically, and making it the responsibility of the hardware interface/driver to convert whatever internal orientation representation it uses to this 'intermediary representation' would seem to make sense.

fmauch:
> High-level question: would it perhaps be better to have the hardware interface offer a quaternion instead of an Euler angle triplet?
> 
> That would seem cleaner to me, in the sense of it being more of an `N-to-1` mathematical abstraction.
> 
> While they aren't necessarily very human friendly, quaternions have definite benefits mathematically, and making it the responsibility of the hardware interface/driver to convert whatever internal orientation representation it uses to this 'intermediary representation' would seem to make sense.

I was discussing that with @RobertWilbrandt, as well and I would agree. But I think that's a question to discuss once the PR on ros2_controllers is there.

gavanderhoorn:
Oh, just noticed #856 actually does it that way.

RobertWilbrandt:
I don't have any strong opinions on this - Going through [REP 103](https://www.ros.org/reps/rep-0103.html#rotation-representation) i went through the options and preferred RPY because the values cannot represent any invalid state (while i would expect that a quaternion should be normalized). From an application point of view i agree, so i changed the representation in pose_broadcaster now.

fmauch:
Binary builds fail as the necessary changes in ros2_control have just been merged into master.

VinDp (approved):
Looks good to me and works as inteded.

Issue #34, opened 2021-01-21
implement velocity_controllers/ScaledJointTrajectoryController


nbbrooks:
Implemented by FZI

PR #1114, opened 2024-09-25
Freedrive Mode

VinDp:
The PR aims at introducing the possibility of enabling the freedrive mode through the ROS2 driver. It relies on the implementation of a specific controller to handle it, in order to correctly deactivate other controllers and avoid unsafe jumps of the robot once the freedrive mode is switched off.

### DIFF
--- a/ur_controllers/include/ur_controllers/freedrive_mode_controller.hpp
+++ b/ur_controllers/include/ur_controllers/freedrive_mode_controller.hpp
@@ -1,0 +1,97 @@
+// Copyright 2023, FZI Forschungszentrum Informatik, Created on behalf of Universal Robots A/S
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the {copyright_holder} nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+//----------------------------------------------------------------------
+/*!\file
+ *
+ * \author  Felix Exner exner@fzi.de
+ * \date    2023-06-29
+ */
+//----------------------------------------------------------------------
+
+#pragma once
+#include <memory>
+#include <tf2_ros/buffer.h>
+#include <tf2_ros/transform_listener.h>
+
+#include <controller_interface/controller_interface.hpp>
+#include <geometry_msgs/msg/pose_stamped.hpp>
+#include <std_msgs/msg/bool.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+namespace ur_controllers
+{
+enum CommandInterfaces
+{
+  FREEDRIVE_MODE_ASYNC_SUCCESS = 25,
+  FREEDRIVE_MODE_CMD = 26,
+};
+enum StateInterfaces
+{
+  INITIALIZED_FLAG = 0u,
+};
+
+class FreedriveModeController : public controller_interface::ControllerInterface
+{
+public:
+  controller_interface::InterfaceConfiguration command_interface_configuration() const override;
+
+  controller_interface::InterfaceConfiguration state_interface_configuration() const override;
+
+  controller_interface::return_type update(const rclcpp::Time& time, const rclcpp::Duration& period) override;
+
+  CallbackReturn on_configure(const rclcpp_lifecycle::State& previous_state) override;
+
+  CallbackReturn on_activate(const rclcpp_lifecycle::State& previous_state) override;
+
+  CallbackReturn on_deactivate(const rclcpp_lifecycle::State& previous_state) override;
+
+  CallbackReturn on_init() override;
+
+private:
+  void readFreedriveModeCmd(const std_msgs::msg::Bool::SharedPtr msg);
+  bool enableFreedriveMode();
+  bool disableFreedriveMode();
+
+  std::unique_ptr<tf2_ros::Buffer> tf_buffer_;
+  std::unique_ptr<tf2_ros::TransformListener> tf_listener_;
+
+  std::shared_ptr<freedrive_mode_controller::ParamListener> param_listener_;
+  freedrive_mode_controller::Params params_;
+
+  std::atomic<bool> freedrive_mode_enable_;
+  std::shared_ptr<rclcpp::Subscription<std_msgs::msg::Bool> freedrive_mode_sub_;
+
+  static constexpr double ASYNC_WAITING = 2.0;
+  /**
+   * @brief wait until a command interface isn't in state ASYNC_WAITING anymore or until the parameter maximum_retries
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Would an action be nicer, semantically?

Start of action -> enable freedrive.

Cancel action -> stop freedrive.

The goal client handle would become something similar to a 'token' or 'mutex' (not exactly, as it wouldn't be as safe or strict, but conceptually).

Concurrent goal -> reject (something/someone else already has 'control' of the freedrive mode).

it would also make it (slightly) harder for something to take over control of the robot by just spamming `Bool` messages.

And it would make it semantically clearer it control some specific functionality (ie: because a custom action goal needs to be submitted, instead of a generic `Bool`).

gavanderhoorn:
One aspect that would need some thought though: time-out (ie: when should freedrive be stopped due to the client no longer 'responding').

There might be some way to exploit liveliness QoS or something, but I'm not sure how that would work with actions.

VinDp:
Currently reached a first version working state, still missing:
- [ ] Handling timeout
- [ ] Tests

Tested on URSim, requires [#34](https://github.com/ros-industrial/ur_msgs/pull/34).

gavanderhoorn:
High-level comment (nice work already though): the (proposed?) action is called `EnableFreeDriveMode`.

To me (non-native speaker though) this implies the action of *enabling* freedrive mode -- so changing the state of the robot from *not* in freedrive mode to having freedrive mode active -- is what this is about.

I realise something like "keep-robot-in-freedrive-mode" is a bit silly as a name, but that seems to be more what this controller does, correct?

fmauch:
> One aspect that would need some thought though: time-out (ie: when should freedrive be stopped due to the client no longer 'responding').
> 
> There might be some way to exploit liveliness QoS or something, but I'm not sure how that would work with actions.

I wrapped my head around this and I currently don't see a way of doing this.

One option would be to use a separate topic for a keepalive signal that could be used opt-in or opt-out, but that feels like a cumbersome workaround. 

> High-level comment (nice work already though): the (proposed?) action is called EnableFreeDriveMode.

I was also stumbling upon that. Maybe "use freedrive mode" could be a proper label? That could imply that, as long as this action is running, the robot's freedrive mode is used. I don't quite like the word "use". Matching synonyms could be "utilize", "operate", "run".

VinDp:
I see the point on the action name and I agree that "enable" is not the best choice to describe what the action does. If we exclude using two words (like StartandKeepFreedriveMode),  I would also choose something similar to "use", and tha regard maybe I like "run" more.

gavanderhoorn:
> I wrapped my head around this and I currently don't see a way of doing this.

I haven't looked into it, but would there be a way to retrieve the liveness QoS data/settings for the feedback or perhaps status publisher and see whether it's still connected to a client?

It might need some unorthodox reaching around/into the action-machinery, but DDS certainly supports these kinds of things.

fmauch:
I've been testing around with that already on https://github.com/fmauch/Universal_Robots_ROS2_Driver/commit/f3a2e3d170e7ab554ff04edd42224c3dd989ce88 (Another feature branch, another action, but I just had that at hand).

Setting a lifespan (which seemed the most obvious thing for me) of 100ms seemed not to have any effect. My test was running a trajectory through our example_move script and then exiting that script during motion.

One question is, whether the underlying action server implementation would care at all, if this would happen. If the message would expire on a DDS level - would that even be visible to the publisher / the action server? Then, API access is a whole other story. 

But thank you for supporting me that I have been poking in the right direction.

gavanderhoorn:
Setting QoS is just one thing that would be needed. I would probably go for Liveliness instead actually.

You'd then have to register a callback to be called whenever the Liveliness requirements are no longer met.

I can't really look for it right now, but I'm pretty sure I've seen discussion/PRs around callbacks for these kinds of events in the ROS 2 stack (perhaps even at RMW level).

fmauch:
I just tested things locally and had discussions with others.

I would like to stick with the heartbeat topic instead of the action as we had it initially for now because of the following reasons:

- Handling timeout using the action doesn't really seem straightforward. From my understanding (and that of my colleagues who know QoS stuff better than I do) it is not intended for a publisher to know whether or not a certain subscriber went away or not.
- Some people were also saying that this is a great misuse of an action. From their point of view an action should be something that's requested for execution, informs the caller about the progress while executing the thing and then finishing at some point either successful or not. Our action here is never supposed to be finished and we don't give the caller any feedback (we could think of some, though). Instead, we want feedback from the caller, whether it is still alive or not.
- With local tests the action wasn't even reliably canceled when using `ros2 action send goal` from the command line. As long as we haven't figured out a reliable way to handle timeout, I don't really want to take that route.
- Using the heartbeat topic should be straightforward to use. Yes, there can be potentially multiple clients requesting to keep freedrive mode active but that's easy to debug with standard ROS tools.
- Others have come up with using the heartbeat topic, as well, e.g. https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/718

I'm not saying that this is the *right* interface to use, but for the sake of getting this PR forward to a mergable state, I would say that another, maybe semantically more meaningful interface can be added at a later stage.

urfeex:
I just resolved the conflicts with the main branch. @VinDp is this ready to review?

ur_controllers/src/freedrive_mode_controller.cpp
```diff
@@ -0,0 +1,330 @@
...
+controller_interface::CallbackReturn
+ur_controllers::FreedriveModeController::on_deactivate(const rclcpp_lifecycle::State&)
+{
+  abort_command_interface_->get().set_value(1.0);
+  freedrive_active_ = false;
+
+  freedrive_sub_timer_.reset();
+  timer_started_ = false;
+
+  stop_logging_thread();
```
urfeex, line 175: This is definitively not RT-safe

ur_controllers/src/freedrive_mode_controller.cpp
```diff
@@ -0,0 +1,330 @@
...
+  return ControllerInterface::on_activate(state);
+}
+
+controller_interface::CallbackReturn
+ur_controllers::FreedriveModeController::on_deactivate(const rclcpp_lifecycle::State&)
+{
+  abort_command_interface_->get().set_value(1.0);
+  freedrive_active_ = false;
+
+  freedrive_sub_timer_.reset();
```
urfeex, line 172: This is probably not RT-safe

ur_controllers/src/freedrive_mode_controller.cpp
```diff
@@ -0,0 +1,330 @@
...
+    timer_started_ = true;
+
+    RCLCPP_INFO(get_node()->get_logger(), "Timer started after receiving first command.");
+  }
+}
+
+void FreedriveModeController::timeout_callback()
+{
+  if (timer_started_ && freedrive_active_) {
+    RCLCPP_INFO(get_node()->get_logger(), "Freedrive mode will be deactivated since client is not reachable.");
```
urfeex, line 264:
```suggestion
    RCLCPP_INFO(get_node()->get_logger(), "Freedrive mode will be deactivated since no new message received.");
```

urfeex (changes requested):
I think, we need to revisit the RT-compatibility, especially regarding `deactivate()`.

ur_controllers/src/freedrive_mode_controller.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+controller_interface::CallbackReturn
+ur_controllers::FreedriveModeController::on_activate(const rclcpp_lifecycle::State& state)
+{
+  change_requested_ = false;
+  freedrive_active_ = false;
+  async_state_ = std::numeric_limits<double>::quiet_NaN();
+
+  first_log_ = false;
+  logging_thread_running_ = false;
```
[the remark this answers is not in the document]
VinDp: True, I forgot to update that. Thanks!

ur_controllers/include/ur_controllers/freedrive_mode_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+  // Command interfaces
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> async_success_command_interface_;
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> enable_command_interface_;
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> abort_command_interface_;
+
+  std::shared_ptr<rclcpp::Subscription<std_msgs::msg::Bool>> enable_freedrive_mode_sub_;
+
+  rclcpp::TimerBase::SharedPtr freedrive_sub_timer_;  ///< Timer to check on the topic
+  mutable std::chrono::seconds timeout_interval_;
+  void readFreedriveModeCmd(const std_msgs::msg::Bool::SharedPtr msg);
```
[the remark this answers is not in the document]
VinDp: Changed with the last commit to `freedrive_cmd_callback`: if you prefer something else feel free to change it, I don't have strong opinions :)

ur_controllers/src/freedrive_mode_controller.cpp
VinDp: True, moved `stop_logging_thread()` to `on_cleanup()`. Thanks!

ur_controllers/src/freedrive_mode_controller.cpp
VinDp: Also moved that to `on_cleanup()` :+1:

ur_controllers/include/ur_controllers/freedrive_mode_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  CallbackReturn on_deactivate(const rclcpp_lifecycle::State& previous_state) override;
+
+  CallbackReturn on_init() override;
+
+private:
+  // Command interfaces
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> async_success_command_interface_;
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> enable_command_interface_;
+  std::optional<std::reference_wrapper<hardware_interface::LoanedCommandInterface>> abort_command_interface_;
```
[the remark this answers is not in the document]
VinDp: I just used the `optional` to avoid to have to initialize the reference. I added a comment in the code to point that out.

urfeex:
@Mergifyio backport humble

PR #1121, opened 2024-10-07
Add passthrough interfaces for joints

fmauch:
As this has been added in the joint description, we need to also parse them from the hardware interface.

This is a part of #944 that is necessary since https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/204 is merged.

This should also fix the current semi-binary failures.


VinDp (approved):
Makes sense and everything seems fine.

PR #1123, opened 2024-10-08
Allow setting the analog output domain when setting an analog output

fmauch:
This basically implements a service for https://github.com/ros-industrial/ur_msgs/pull/30


VinDp (approved):
Works as intended and looks good to me.

PR #1124, opened 2024-10-08
Update ur_msgs branch for upstream workspaces

fmauch:
Also removed repos files of old distros, as they are being used from their respective branches.

This is a requirement for #1123


VinDp (approved):
Looks good to me!

Issue #1129, opened 2024-10-08
Scaled_joint_trajectory_controller does not respond to speed change on teach pendant


Adrian2702:
### Affected ROS2 Driver version(s)

Humble

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

Build both the ROS driver and UR Client Library from source

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

5.13.0

### How is the ROS driver used.

Through the robot teach pendant using External Control URCap

### Issue details

# Summary
Controlling a UR10e robot with the URCap “External-Control” from an external PC. Waypoints are sent from the PC to the real robot, which it also follows, but the movement is aborted due to a time-out error as soon as the speed slider is not at 100%

## Issue details
We are trying to control a UR10e robot with Polyscope 5.13 via ROS2 Humble. The robot is connected to an external PC via Ethernet. The URCap “External-Control” is installed on the real robot. When I press the play button, I get the following output on the screen:
```
[ros2_control_node-1] [INFO] [1728380483.965475739] [UR_Client_Library:]: Robot requested program
[ros2_control_node-1] [INFO] [1728380483.965620055] [UR_Client_Library:]: Sent program to robot
[ros2_control_node-1] [INFO] [1728380484.260967978] [UR_Client_Library:]: Robot connected to reverse interface. Ready to receive control commands.
```
The Scaled_joint_trajectory_controller is active and receives waypoints via a C++ script. Everything works well and the robot follows the path. However, this is only possible if the speed controller is set to 100%. If I slow it down, I get the following error:
```
MoveGroupInterface::execute() failed or timeout reached
```
The robot moves to the first points of the path and then aborts the movement with the MoveGroup Error.

## Steps to Reproduce
Start the Launch-Files``` ros2 launch ur_robot_driver ur_control.launch.py ur_type:=ur10e robot_ip:=192.168.1.102 launch_rviz:=false ``` and ```ros2 launch ur_moveit_config ur_moveit.launch.py ur_type:=ur10e```

## Expected Behavior
The robot will slow down his movements, when the speed slider on the teach pendant is scaled down but don't aborted the movement.

## Actual Behavior
The robot moves to the first few points and then aborted the movement.

# Workaround Suggestion
[...]

PR #1132, opened 2024-10-08
[moveit] Add config for trajectory execution and disable execution mo…

fmauch:
…nitoring by default

With the scaled JTC execution time monitoring doesn't make much sense as long as MoveIt is not aware of the trajectory scaling.

### DIFF
--- a/ur_moveit_config/config/moveit_controllers.yaml
+++ b/ur_moveit_config/config/moveit_controllers.yaml
@@ -1,6 +1,12 @@
 # MoveIt uses this configuration for controller management
 moveit_controller_manager: moveit_simple_controller_manager/MoveItSimpleControllerManager
 
+trajectory_execution:
+  allowed_execution_duration_scaling: 1.2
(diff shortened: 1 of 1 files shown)


VinDp (approved):
Tested: with this the execution is not stopped anymore when the velocity is not 100%, so it definitely solves [#1129](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1129) for me.

gavanderhoorn:
I would strongly recommend to document the change in this PR, perhaps by adding a warning in a (very) visible place.

It deviates from the default MoveIt behaviour (ie: monitoring trajectory execution) -- and I understand why -- but it will also mean any mistakes or other causes for trajectory execution running 'over time' will not be handled by MoveIt's TEM anymore.

This now delegates responsibility for keeping track of trajectory execution to essentially action clients, which could result in unsuspecting clients running into 'hangups' of MoveIt (as it can't abort goals automatically anymore).

Issue #1135, opened 2024-10-09
Document that (and why) we disable MoveIt trajectory execution monitoring


fmauch:
### Feature summary

In #1132 we disabled trajectory execution monitoring (TEM) because it doesn't make much sense to use that with the scaled JTC.

However, as pointed out in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1132#issuecomment-2401892674 this might lead to unexpected behavior for users that are familiar with TEM as this basically shifts the responsibility to monitor that to the user (the action client).

I think, we should at least add that to the [usage section](https://docs.universal-robots.com/Universal_Robots_ROS2_Documentation/doc/ur_robot_driver/ur_robot_driver/doc/usage/move.html#using-moveit) and the [package doc](https://docs.universal-robots.com/Universal_Robots_ROS2_Documentation/doc/ur_robot_driver/ur_moveit_config/doc/index.html#ur-moveit-config).

PR #1136, opened 2024-10-09
Add note about TEM

fmauch:
This implements #1135

### DIFF
--- a/ur_moveit_config/doc/index.rst
+++ b/ur_moveit_config/doc/index.rst
@@ -27,8 +27,8 @@ robot as explained `here <https://moveit.picknik.ai/main/doc/tutorials/quickstar
 .. note::
    Usually, MoveIt uses trajectory execution monitoring (TEM). If executing a trajectory takes too long,
    e.g. because the action server died or the robot is blocked, this will lead to an error in
-   trajectory execution. However, with the scaled joint trajectory controller offered by this driver,
-   this doesn't make much sense (as long as TEM is not aware of the scaling), as execution time can
+   trajectory execution. However, this isn't necessarily compatible with the scaled joint
+   trajectory controller (as long as TEM is not aware of the scaling), as execution time can
    be unbounded on purpose. Hence, TEM is disabled in the example MoveIt configuration:
 
    .. literalinclude:: ../config/moveit_controllers.yaml
--- a/ur_moveit_config/doc/index.rst
+++ b/ur_moveit_config/doc/index.rst
@@ -23,3 +23,16 @@ interaction using
 
 Now you should be able to use the MoveIt Plugin in rviz2 to plan and execute trajectories with the
 robot as explained `here <https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>`_.
+
+.. note::
+   Usually, MoveIt uses trajectory execution monitoring (TEM). If executing a trajectory takes too long,
+   e.g. because the action server died or the robot is blocked, this will lead to an error in
+   trajectory execution. However, with the scaled joint trajectory controller used by this driver,
+   this doesn't make much sense (as long as TEM is not aware of the scaling), as execution time can
+   be unbounded on purpose. Hence, TEM is disabled in the example moveit configuration:
+
+   .. literalinclude:: ../config/moveit_controllers.yaml
+      :language: yaml
+      :start-at: trajectory_execution:
+      :end-at: execution_duration_monitoring
+      :caption: moveit_controllers.yaml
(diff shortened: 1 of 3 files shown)


VinDp (approved):
Looks good to me.

ur_moveit_config/doc/index.rst
```diff
@@ -23,3 +23,16 @@ interaction using
 
 Now you should be able to use the MoveIt Plugin in rviz2 to plan and execute trajectories with the
 robot as explained `here <https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>`_.
+
+.. note::
+   Usually, MoveIt uses trajectory execution monitoring (TEM). If executing a trajectory takes too long,
+   e.g. because the action server died or the robot is blocked, this will lead to an error in
+   trajectory execution. However, with the scaled joint trajectory controller used by this driver,
+   this doesn't make much sense (as long as TEM is not aware of the scaling), as execution time can
+   be unbounded on purpose. Hence, TEM is disabled in the example moveit configuration:
```
gavanderhoorn, line 32:
```suggestion
   be unbounded on purpose. Hence, TEM is disabled in the example MoveIt configuration:
```

ur_moveit_config/doc/index.rst
```diff
@@ -23,3 +23,16 @@ interaction using
 
 Now you should be able to use the MoveIt Plugin in rviz2 to plan and execute trajectories with the
 robot as explained `here <https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>`_.
+
+.. note::
+   Usually, MoveIt uses trajectory execution monitoring (TEM). If executing a trajectory takes too long,
+   e.g. because the action server died or the robot is blocked, this will lead to an error in
+   trajectory execution. However, with the scaled joint trajectory controller used by this driver,
```
gavanderhoorn, line 30:
```suggestion
   trajectory execution. However, with the scaled joint trajectory controller offered by this driver,
```

ur_moveit_config/doc/index.rst
```diff
@@ -23,3 +23,16 @@ interaction using
...
+   this doesn't make much sense (as long as TEM is not aware of the scaling), as execution time can
```
gavanderhoorn, line 31: Perhaps "does not work" or "is incompatible with".

It still makes sense, it's just that with the current implementation on MoveIt's side it will result in undesirable behaviour, as it always sees execution time violations.

fmauch: I did not want to place TEM as a bad idea or anything, so "doesn't make sense" isn't the best wording, yes. "does not work" also isn't entirely true, as it does work in terms that is monitors whether the trajectory is executed in the expected time. Howabout "isn't necessarily compatible with"? I mean, it is pretty compatible with it, as long as no scaling is happening.

ur_moveit_config/doc/index.rst
```diff
@@ -23,3 +23,16 @@ interaction using
 
 Now you should be able to use the MoveIt Plugin in rviz2 to plan and execute trajectories with the
 robot as explained `here <https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>`_.
+
+.. note::
+   Usually, MoveIt uses trajectory execution monitoring (TEM). If executing a trajectory takes too long,
+   e.g. because the action server died or the robot is blocked, this will lead to an error in
+   trajectory execution. However, this isn't necessarily compatible with the scaled joint
+   trajectory controller (as long as TEM is not aware of the scaling), as execution time can
+   be unbounded on purpose. Hence, TEM is disabled in the example MoveIt configuration:
```
gavanderhoorn, line 32: Another suggestion:

```suggestion
The MoveIt configuration provided here has Trajectory Execution Monitoring *disabled*, as the Scaled Joint Trajectory Controller may cause trajectories to be executed at a lower velocity than they were originally planned by MoveIt. MoveIt's TEM however is not aware of this deliberate slow-down due to scaling and will in most cases unnecessarily (and unexpectedly) abort goals.

Until this incompatibility is resolved, the default value for `execution_duration_monitoring` is set to `false`. Users who wish to temporarily (re)enable TEM at runtime (for use with other, non-scaling controllers) can do so using the ROS 2 parameter services supported by MoveIt.
```

a bit more text, but perhaps more specific about what's going on, what would happen if it wasn't disabled, which parameter is specifically responsible for disabling it and how users could (temporarily perhaps) re-enable it at runtime.

fmauch: That sounds good, thank you!

PR #1138, opened 2024-10-10
Fixed dependency for ur_msgs from upstream

pucciland95:
## Problem
As of today, 10/10/24, it is **impossible** to binary build this repo using `src/Universal_Robots_ROS2_Driver/Universal_Robots_ROS2_Driver-not-released.<ros-distro>.repos`.

Also your CI [shows the problem](https://github.com/pucciland95/Universal_Robots_ROS2_Driver/blob/main/ci_status.md).
It issues and error since it fails to get the latest `ur_msgs` changes. Indeed, when getting the latest binary of `ur_msgs` from rosdep it pulls version 2.0.0 despite the latest is version 2.0.1. I don't get why honestly...

## Proposed solution
As a temporary fix, It is necessary to clone `ur_msgs` from upstream.

### DIFF
--- a/Universal_Robots_ROS2_Driver-not-released.humble.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.humble.repos
@@ -8,3 +8,7 @@ repositories:
     type: git
     url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Description.git
     version: humble
+  ur_msgs:
+    type: git
+    url: https://github.com/ros-industrial/ur_msgs.git
+    version: humble-devel
--- a/Universal_Robots_ROS2_Driver-not-released.humble.repos
+++ b/Universal_Robots_ROS2_Driver-not-released.humble.repos
@@ -11,4 +11,4 @@ repositories:
   ur_msgs:
     type: git
     url: https://github.com/ros-industrial/ur_msgs.git
-    version: humble-devel
+    version: humble


gavanderhoorn:
> I don't get why honestly...

because even though `2.0.1` has been released, it hasn't been synced yet to the main package repositories. See [status_page/ros_humble_default.html?q=ur_msgs](https://repo.ros2.org/status_page/ros_humble_default.html?q=ur_msgs).

So this will work again after that has happened, or if you configure the [testing repository](https://docs.ros.org/en/humble/Installation/Testing.html).

pucciland95:
Hi, @gavanderhoorn; thank you for answering the question and clarifying the issue.

Could you please explain to me more about how this works? You released a new version of `ur_msgs` on Git Hub and then what happens?

gavanderhoorn:
I believe this is actually quite well described by [deb testing repository](https://docs.ros.org/en/humble/Installation/Testing.html#deb-testing-repository) (emphasis mine):

> When packages are released into a ROS distribution (using bloom), the buildfarm builds them into deb packages which are stored temporarily in the `building` apt repository. As dependent packages are rebuilt, an automatic process periodically synchronizes the packages in `building` to a secondary repository called `ros-testing`. `ros-testing` is intended as a soaking area where developers and bleeding-edge users may give the packages extra testing, before they are manually synced into the public ros repository from which users typically install packages.
>
> **Approximately every two weeks, the rosdistro’s release manager manually synchronizes the contents of `ros-testing` into the `main` ROS repository.**

that's a terse description of what happens after a ROS package developer/maintainer 'releases' their package.

fmauch:
Please see point 6 of https://docs.universal-robots.com/Universal_Robots_ROS2_Documentation/doc/ur_robot_driver/ur_robot_driver/doc/installation/installation.html#build-from-source (Though that link to the build status should be fixed)

We are currently introducing some breaking changes for rolling and hence it is expected that the binary-main builds fail until that state has been synchronized to the main repo. That's why we don't suggest building the driver from source unless needed. If needed, the `Universal_Robots_ROS2_Driver.<ros-distro>.repos` files should be used as upstream workspace. The `non-released` files are a leftover from the time when not all upstream dependencies were actually released (pre-foxy). I was already considering deleting them at some point...

pucciland95:
Ok, clear,

Indeed I followed the tutorial you linked to build from source the code and it was not working. 
What should I do then with this pull request? Should I close it?

Cheers and thank's again for the explanation

fmauch:
I'm going to close this as explained above this is not really desired.

PR #1137, opened 2024-10-09
Readme update hosted docs

fmauch:
Update broken links from README and add a URL checker to CI.

PR #1142, opened 2024-10-11
Disable link checks for architecture_coarse.svg

fmauch:
When adding the link checks in #1137, this actually slipped through.


VinDp (approved):
Looks good now.

PR #1144, opened 2024-10-12
Pass `use_sim_time` to MoveIt's RViz instance

fmauch:
When running with sim_time, we need to pass that to the RViz instance, as well.

This is a part of fixing https://github.com/UniversalRobots/Universal_Robots_ROS2_GZ_Simulation/issues/57


VinDp (approved):
Looks good to me and solves the addressed issue.

PR #1151, opened 2024-10-14
Revert "Add passthrough interfaces for joints (#1121)"

fmauch:
This reverts commit 3068607b1de03a6e6a4687a3da51b694449794a6.

Complementing https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/214

I had a bad feeling about this and hence also only released it to rolling. Thinking about this a bit, it occurred to me that -- since we want to send other motion types in the future -- it doesn't make sense necessarily to add this to the joints.


VinDp (approved):
Looks good to me.

PR #1098, opened 2024-09-11
Fix component lifecycle

fmauch:
This is closing #1097 

As mentioned in the issue I would like to combine this issue with tests for lifecycle management.

Issue #1097, opened 2024-09-09
Failure During Hardware Component Reconfiguration


sahand-ghaffari-ocado:
### Affected ROS2 Driver version(s)

2.2.14

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

Build both the ROS driver and UR Client Library from source

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

5.17.0

### How is the ROS driver used.

Headless without using the teach pendant

# Problem Description

Reconfiguring the UR hardware component (which can be done using the `~/set_hardware_component_state` service call to unconfigure and then activate the UR hardware component) results in a code failure due to following issues

1. The `registerUrclLogHandler()` [function](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/humble/ur_robot_driver/src/urcl_log_handler.cpp#L82) which is called in the  `on_configure()` [function](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/humble/ur_robot_driver/src/hardware_interface.cpp#L421) uses `std::move` to transfer ownership of `g_log_handler`. As a result, calling the `on_configure() `function again causes a failure. This is because the `g_log_handler` pointer has already been moved, and it no longer has access to the `setTFPrefix()` function.
2. The `checkAsyncIO()` [function](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/humble/ur_robot_driver/src/hardware_interface.cpp#L518) cannot be called in the` asyncThread() `function after reconfiguring the hardware component. This is because the `async_thread_shutdown_` variable is set to true in the `on_cleanup()` [function](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/humble/ur_robot_driver/src/hardware_interface.cpp#L477) and is not reset to false in the `on_configure()` function.

# Workaround Suggestion
To avoid such crashes, following changes are suggested

1. It is suggested to declare and assign the pointer within the `registerUrclLogHandler()` function instead of doing so globally. Therefore, it is suggested to modify the function as follows:

```
void registerUrclLogHandler(const std::string& tf_prefix)
{
  if (g_registered == false) {
    std::unique_ptr<UrclLogHandler> g_log_handler(new UrclLogHandler);
[...]

PR #1154, opened 2024-10-14
fix: recreate urcl log handler pointer if necessary

domire8:
We have discovered an [issue](https://github.com/aica-technology/api/issues/187) with the `UrclLogHandler` that occurs when a resource manager is creating the UR hardware interface twice in a row without stopping ROS in between. I'm not sure if that can be recreated on your side easily but you could try:
1. create controller manager for UR
2. destroy controller manager while keeping ros2 alive
3. recreate controller manager for UR, this will result in a segmentation fault

You will realize that the `g_log_handler` pointer has in fact been reset and not recreated properly, since it's initialization is to be found in the source file. My null pointer check should avoid that segfault.

Side question, what was the design intent of the `UrclLogHandler`? Were you going for a Singleton? That influences maybe how this problem could be fixed.

### DIFF
--- a/ur_robot_driver/src/urcl_log_handler.cpp
+++ b/ur_robot_driver/src/urcl_log_handler.cpp
@@ -82,6 +82,9 @@ void UrclLogHandler::log(const char* file, int line, urcl::LogLevel loglevel, co
 void registerUrclLogHandler(const std::string& tf_prefix)
 {
   if (g_registered == false) {
+    if (g_log_handler == nullptr) {
+      g_log_handler = std::make_unique<UrclLogHandler>();
+    }
     g_log_handler->setTFPrefix(tf_prefix);
     // Log level is decided by ROS2 log level
     urcl::setLogLevel(urcl::LogLevel::DEBUG);


fmauch:
Thank you for pointing that out. I am not sure if I understand everything, you are writing:

> destroy controller manager while keeping ros2 alive

What exactly do you mean by "keeping ros2 alive"?

Anyway, this is a valid concern and in the back of my head that rang a bell, that we did touch this in the past: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1098

I don't quite understand why past-me came to the conclusion that it would be a good idea not to include the fix in order to write tests, though. Since #1098 also solves another issues about lifecycle state changing: Would you mind if we merged that instead?

domire8:
> What exactly do you mean by "keeping ros2 alive"?

Yeah I'm sorry I got so used to our framework that I have a hard time describing issues in native ROS 2. We use dynamic composition a lot and that means that our supervisor process might load a controller manager with a UR and then unload when the user requests it and then load it again. When I said "keep ros2 alive" I meant that there are a bunch of other nodes still alive while the hardware interface is stopped and removed and then added again.

> Would you mind if we merged that instead?

The issue is indeed well described in #1097 and #1098 should be a valid fix too, so yes we can close this PR without merge. I'll find another opportunity to contribute :smile:

fmauch:
Closing, now that #1098 is merged. Thank you @domire8 for bringing that up again!

PR #1175, opened 2024-11-05
Add launch argument to optionally omit the controller_stopper node

eholum:
Adds an argument to [ur_control.launch.py](https://github.com/eholum/Universal_Robots_ROS2_Driver/blob/controller-stopper-option/ur_robot_driver/launch/ur_control.launch.py#L476) to optionally launch the `controller_stopper` node.

In our hardware setup we have multiple types of robots (URs, rails, lifts, AMRs, etc), and have implemented some custom software e-stops for halting everything in the system, then restarting controllers as needed. While we normally rely on the `controller_stopper`, we've found that it steps on the toes of our custom e-stopper. It would be handy to optionally not include the node in a way that lets us continue to use the driver's launch files as is.

I don't imagine this change will be something that most UR users will need. That said, it may be a nice to have if anyone runs into a similar situation as we are.

Testable with variations of:

```
ros2 launch ur_robot_driver ur5e.launch.py robot_ip:=0.0.0.0 use_mock_hardware:=false use_controller_stopper:=false
```

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -66,6 +66,7 @@ def launch_setup(context):
     use_tool_communication = LaunchConfiguration("use_tool_communication")
     tool_device_name = LaunchConfiguration("tool_device_name")
     tool_tcp_port = LaunchConfiguration("tool_tcp_port")
+    use_controller_stopper = LaunchConfiguration("use_controller_stopper")
 
     control_node = Node(
         package="controller_manager",
@@ -120,7 +121,9 @@ def launch_setup(context):
         name="controller_stopper",
         output="screen",
         emulate_tty=True,
-        condition=UnlessCondition(use_mock_hardware),
+        condition=IfCondition(
+            AndSubstitution(use_controller_stopper, NotSubstitution(use_mock_hardware))
+        ),
         parameters=[
             {"headless_mode": headless_mode},
             {"joint_controller_active": activate_joint_controller},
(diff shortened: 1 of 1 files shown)


ur_robot_driver/launch/ur_control.launch.py
```diff
@@ -120,14 +121,16 @@ def launch_setup(context):
...
+            AndSubstitution(use_controller_stopper, NotSubstitution(use_mock_hardware))
+        ),
         parameters=[
             {"headless_mode": headless_mode},
             {"joint_controller_active": activate_joint_controller},
             {
                 "consistent_controllers": [
                     "io_and_status_controller",
-                    "force_torque_sensor_broadcaster",
+                    "force_torque_sensoqr_broadcaster",
```
eholum, line 133:
```suggestion
                    "force_torque_sensor_broadcaster",
```

urfeex:
I would actually rather not have this. The controller_stopper is an essential piece in a) avoiding sudden motions of the robot in case the user has sent a target while the robot program was not running and b) help making clear that the robot currently cannot accept new targets since the controller is not running.

What is your reasoning why you would not to have it running? Anyway, users can still build their own launchfiles without the controller stopper if they really want to. That obviously requires a bit more understanding than just running a launch file, but disabling the controller stopper also requires understanding what its purpose is and what I as a user have to take care of myself.

eholum:
Thanks for the quick feedback, @urfeex! That makes complete sense to me, we'll stick with building our own launch files.

We run into issues when using applications that don't necessary reset their controller commands after something is paused. For example, MoveIt Servo may continue sending commands to a JTC even after the JTC is stopped by the controller stopper. When the UR is reset and the controller reactivated there will be a resulting joint jump. 

We have similar situations with other types of hardware where the application or controller needs to be reset before a controller is re-activated, so to handle that we've rolled everything into a software e-stop. What we want on top of that is a snapshot of the controller manager's state prior to triggering a stop, which we _might_ not be able to get if the UR's controller stopper triggers before our node's does.

In any case, thanks for the information. If this option were there I think it would be more trivial for a novice user to do something dangerous, whereas it's a much more deliberate choice on our end. 

Closing this.

PR #1176, opened 2024-11-06
Update Humble upstream workspace

urfeex:
Update the upstream workspace regarding

- Add realtime_tools from ros-controls
- Update branch of kinematics_interface to humble as the master version got API incompatible
- Make upstream ws for binary build empty


VinDp (approved):
Good fix!

PR #1189, opened 2024-11-23
Update RealTimeBox

christophfroehlich:
The deprecation notice did not work because the header file changed #fail

https://github.com/ros-controls/realtime_tools/pull/146


urfeex:
Thank you for catching up with this! It looks like CI has been started before the merge was done in realtime tools. I'll restart builds and see again.

christophfroehlich:
Exactly, I just realized that afterwards that ros2_controllers repo is broken as well. 
The master branch of this repo should be fine since [the last PR in realtime_tools](https://github.com/ros-controls/realtime_tools/pull/202), maybe you want to wait for merging this until a new version of realtime_tools got released.

urfeex:
I rather have working semi-binary builds than binary-testing builds as binary-testing is also tested by the buildfarm. Since without this merged, all semi-binary builds for jazzy and rolling break, I will merge this.

urfeex (approved):
Things look  good and semi-binary builds are successful.

Thanks @christophfroehlich for directly addressing this!

PR #1197, opened 2024-12-02
Initialize segments in constructor of DHRobot in calibration.hpp

b-bremer:
Also add missing #include <cassert>


urfeex (approved):
This send line a reasonable change.

PR #1208, opened 2024-12-16
Add control_toolbox to semi-binary repos files

christophfroehlich:
Currently, the semi-binary jobs are failing because ros2_controllers need a new class from control_toolbox.


urfeex (approved):
Thank you for taking the time to think of us :-)

Issue #799, opened 2023-09-13
Improve tutorial on calibration extraction


fmauch:
There are a couple of things that can be improved:

- During the extraction make sure that the robot is actually switched on (at least in IDLE)
- Explain how to verify that the correct calibration is actually being used
- Explain how to properly use it (create an own description / launch package)

Related issues:
- https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/711
- https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/788
- https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/791

cpsiff:
Hello, I'd like to help with this but I'm not sure where I can make the changes to make a pull request. When I click "edit on GitHub" on [this link](https://docs.universal-robots.com/Universal_Robots_ROS2_Documentation/doc/ur_robot_driver/ur_robot_driver/doc/installation/robot_setup.html#extract-calibration-information) I get a 404 error (directed [here](https://github.com/UniversalRobots/Universal_Robots_ROS2_Documentation/blob/main//doc/ur_robot_driver/ur_robot_driver/doc/installation/robot_setup.rst)). Can you direct me to where I can make edits to that doc?

I was thinking at the very least it would be helpful to add documentation to the bottom of [this page](https://docs.universal-robots.com/Universal_Robots_ROS2_Documentation/doc/ur_robot_driver/ur_robot_driver/doc/installation/robot_setup.html#extract-calibration-information) describing how to pass the extracted calibration as a parameter to ur_robot_driver (as explained in #918) and how to verify the checksum by searching the command line output. AFAIK the only way to know to do these things currently is by reading the API or looking through Github issues.

urfeex:
Hi @cpsiff thank you for reaching out, any help would be very appreciated! Sources for the page you referred to should be that: 
https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/doc/installation/robot_setup.rst

While being in there, we could add a `:github_url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/doc/installation/robot_setup.rst` at the top in order to fix that edit link. That should, however probably be done for every page, so that might be a good task for a separate PR.

To build the complete documentation, please see https://github.com/UniversalRobots/Universal_Robots_ROS_Documentation

PR #1214, opened 2024-12-18
improve docs around usage of extracted calibration info

cpsiff:
Added to / modified documentation to make it more clear how to utilize extracted calibration information, and how to verify that the information is being utilized. Also clarified that the robot must be on to extract calibration info.

I added a description of the parameter-passing method, but not the "create your own description package and use that" method. That seemed a little too in-depth and unnecessary for the getting started documentation, when the parameter method will work for most cases. But if I'm missing something, maybe we want to describe the second method as well.

Related issues: #799

### DIFF
--- a/ur_robot_driver/doc/usage/startup.rst
+++ b/ur_robot_driver/doc/usage/startup.rst
@@ -28,6 +28,7 @@ Allowed ``ur_type`` strings: ``ur3``, ``ur3e``, ``ur5``, ``ur5e``, ``ur10``, ``u
 Other important arguments are:
 
 
+* ``kinematics_params_file`` (default: *None*) - Path to the calibration file extracted from the robot, as described in :ref:`calibration_extraction`.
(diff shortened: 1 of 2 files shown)


ur_robot_driver/doc/usage/startup.rst
```diff
@@ -72,6 +73,23 @@ Depending on the :ref:`robot control mode<operation_modes>` do the following:
 
 .. _continuation_after_interruptions:
 
+Verify calibration info is being used correctly
```
urfeex, line 76: When using a wrong calibration, the driver will output an error. I always have stdout coloring enabled for ROS, so I usually look for a red error message during startup to verify that I have specified the correct calibration. Maybe we should mention that?

cpsiff: Sure. added in [8437787](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1214/commits/84377872491216a4220a50cc602610ce4de7c746).

urfeex (approved):
This is looking good now, thank you for your contribution @cpsiff

PR #1215, opened 2024-12-21
Add missing test dependencies for ur_controllers

urfeex:
Currently, our buildfarm builds fail for Humble, but it should also be fixed for the main branch.


fmauch (approved):
Unorthodoxly approving this myself due to holiday season but a broken buildfarm build.

PR #1226, opened 2025-01-09
Running integration tests with mock hardware

URJala:
Added some launch arguments and some checks to avoid connecting to the dashboard interface when using mock hardware, as that will not work. Maybe not the most elegant, but it works for now. Currently passthrough controller does not work at all with mock hardware, and is bypassed if using that in the test. test_trajectory_scaled_aborts_on_violation fails, as the hardware doesnt abort. test_set_io also fails as the controller cant verify that a pin has been set.


urfeex (commented):
I think it is a good idea to split up the robot_driver test into individual tests.

However, I think it would be beneficial to be able to specify mock hardware outside of the test parametrization. This way, we can run the mock hardware tests also when the integration tests flag is not set. This would effectively allow running some tests on the buildfarm, as well.

ur_robot_driver/test/test_common.py
```diff
@@ -364,10 +366,47 @@ def generate_dashboard_test_description():
     )
 
 
+def generate_mock_hardware_test_description(
```
urfeex, line 369: I think that doesn't require a `use_mock_hardware` parameter, right?

ur_robot_driver/test/test_common.py
```diff
@@ -364,10 +366,47 @@ def generate_dashboard_test_description():
...
+    )
+
+    return LaunchDescription(_declare_launch_arguments() + [ReadyToTest(), robot_driver])
+
+
 def generate_driver_test_description(
     tf_prefix="",
     initial_joint_controller="scaled_joint_trajectory_controller",
     controller_spawner_timeout=TIMEOUT_WAIT_SERVICE_INITIAL,
+    use_mock_hardware="false",
```
urfeex, line 409: Since you created a separate one, we don't need to add the parameter here.

ur_robot_driver/urdf/ur.ros2_control.xacro
```diff
@@ -300,7 +300,9 @@
 
 
       <gpio name="${tf_prefix}get_robot_software_version">
-        <state_interface name="get_version_major"/>
+        <state_interface name="get_version_major">
+          <param name="initial_value">1</param>
```
urfeex, line 304: in the config controller we test that it is not equal to 0. With this change, that would always be true, right?

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+    ConfigurationInterface,
+    generate_mock_hardware_test_description,
+    ROBOT_JOINTS,
+    TIMEOUT_EXECUTE_TRAJECTORY,
+)
+
+
+@pytest.mark.launch_test
+@launch_testing.parametrize("tf_prefix, use_mock_hardware", [("", "true"), ("my_ur_", "true")])
+def generate_test_description(tf_prefix, use_mock_hardware):
```
urfeex, line 60: I guess, the `use_mock_hardware` parameter can be removed from that.

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class RobotDriverTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls, use_mock_hardware):
+        # Initialize the ROS context
+        rclpy.init()
+        cls.node = Node("mock_hardware_test")
+        time.sleep(1)
+        cls.mock_hardware = use_mock_hardware == "true"
```
urfeex, line 73:
```suggestion
```

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+            self._configuration_controller_interface.get_robot_software_version().major, 1
+        )
+
+    def test_start_scaled_jtc_controller(self):
+        self.assertTrue(
+            self._controller_manager_interface.switch_controller(
+                strictness=SwitchController.Request.BEST_EFFORT,
+                activate_controllers=["scaled_joint_trajectory_controller"],
+            ).ok
+        )
```
urfeex, line 113: When BEST_EFFORT is used, that will return ok, as soon as the controller exists.

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+
+        self._scaled_follow_joint_trajectory = ActionInterface(
+            self.node,
+            "/scaled_joint_trajectory_controller/follow_joint_trajectory",
+            FollowJointTrajectory,
+        )
+
+    def setUp(self):
+        time.sleep(1)
+        self.assertTrue(self._io_status_controller_interface.resend_robot_program().success)
```
urfeex, line 96: I don't think that this makes a lot of sense here.

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+        logging.info("Sending goal for robot to follow")
+        goal_handle = self._scaled_follow_joint_trajectory.send_goal(trajectory=trajectory)
+        self.assertTrue(goal_handle.accepted)
+
+        # Verify execution
+        result = self._scaled_follow_joint_trajectory.get_result(
+            goal_handle,
+            TIMEOUT_EXECUTE_TRAJECTORY,
+        )
+        self.assertEqual(result.error_code, FollowJointTrajectory.Result.SUCCESSFUL)
```
urfeex, line 198: That test doesn't make a lot of sense, since the mock hardware isn't scaling.

ur_robot_driver/test/test_mock_hardware.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def test_start_scaled_jtc_controller(self):
+        self.assertTrue(
+            self._controller_manager_interface.switch_controller(
+                strictness=SwitchController.Request.BEST_EFFORT,
+                activate_controllers=["scaled_joint_trajectory_controller"],
+            ).ok
+        )
+
+    def test_trajectory(self, tf_prefix):
```
urfeex, line 115: These are basically the same as the ones from the test_scaled_jtc.py test. I think, we should share the code, then.

urfeex (changes requested):

ur_robot_driver/test/integration_test_io_controller.py
urfeex, line 1: Shouldn't this test also be able to run on mock hardware? I see, it doesn't, but I don't see a conceptual reason why it shouldn't.

ur_robot_driver/test/integration_test_config_controller.py
```diff
@@ -0,0 +1,89 @@
...
+)
+
+
+@pytest.mark.launch_test
+@launch_testing.parametrize("tf_prefix", [(""), ("my_ur_")])
+def generate_test_description(tf_prefix):
+    return generate_driver_test_description(tf_prefix=tf_prefix)
+
+
+class RobotDriverTest(unittest.TestCase):
```
urfeex, line 56:
```suggestion
class ConfigControllerTest(unittest.TestCase):
```

ur_robot_driver/test/integration_test_config_controller.py
```diff
@@ -0,0 +1,89 @@
...
+    @classmethod
+    def setUpClass(cls):
+        # Initialize the ROS context
+        rclpy.init()
+        cls.node = Node("robot_driver_test")
```
urfeex, line 61:
```suggestion
        cls.node = Node("config_controller_test")
```

urfeex (changes requested):
This seems fine so far and I think with renaming all the tests I would be fine with merging this. There is more potential for refactoring things like the initialization, but for this PR I think it's fine to leave it like that.

ur_robot_driver/test/integration_test_io_controller.py
URJala: There is no real reason that it shouldn't, but with the way the controller sets io's, it apparently cannot verify if a pin was set, and fails the test because the pins indeed end up at the wrong value. The controller also prints a little warning that this can happen with mock hardware, so I don't think it should be tested, as it doesn't behave correctly.

urfeex (approved):

PR #1227, opened 2025-01-09
Add information about which driver features dont work with mock hardw…

URJala:
…are.

### DIFF
--- a/ur_robot_driver/doc/usage/simulation.rst
+++ b/ur_robot_driver/doc/usage/simulation.rst
@@ -43,3 +43,9 @@ environment for testing of "piping" of hardware and controllers, as well as test
 descriptions. For more details see `ros2_control documentation
 <https://control.ros.org/rolling/doc/ros2_control/hardware_interface/doc/mock_components_userdoc.html>`_
 for more details.
+Some driver functions don't work with mock hardware currently:
+
+* The TCP pose broadcaster does not work.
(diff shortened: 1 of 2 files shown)


ur_robot_driver/doc/usage/simulation.rst
```diff
@@ -43,3 +43,9 @@ environment for testing of "piping" of hardware and controllers, as well as test
 descriptions. For more details see `ros2_control documentation
 <https://control.ros.org/rolling/doc/ros2_control/hardware_interface/doc/mock_components_userdoc.html>`_
 for more details.
+Some driver functions don't work with mock hardware currently:
+
+* The TCP pose broadcaster does not work.
+* The passthrough trajectory controller does not function when calling the follow joint trajectory action.
+* The force mode controller also does not respond when trying to start force mode.
+* The GPIO controller cannot verify that it has changed the state of an I/O pin, so it will report a failure when trying to set an I/O pin.
```
urfeex, line 51:
```suggestion
.. note::
   Some driver functionalities currently don't work with mock hardware:

   * The TCP pose broadcaster does not work.
   * The passthrough trajectory controller does not function when calling the follow joint trajectory action.
   * The force mode controller also does not respond when trying to start force mode.
   * The GPIO controller cannot verify that it has changed the state of an I/O pin, so it will report a failure when trying to set an I/O pin.
```

urfeex (approved):
Thanks for adding this information!

PR #1239, opened 2025-01-17
Remove Iron from the active distributions

urfeex:
As Iron is EOL since end of November we should clearly state that here, as well.


urrsk (approved):
We should also clean up the builds for Foxy and Galactic as well

urfeex:
> We should also clean up the builds for Foxy and Galactic as well

Done in 8bf8838

PR #1240, opened 2025-01-20
fix sphinx doc link in ur_robot_driver

urrsk:
Fix sphinx doc link in ur_robot_driver to make the pipeline green again


urfeex (approved):
Thanks for addressing that. This should probably go to other repos, as well.

PR #1234, opened 2025-01-14
Remove build warnings for Humble

urfeex:
This PR aims at removing all build warnings for Humble.

Waiting for realtime-tools to be released before this is ready-to-review.

PR #1241, opened 2025-01-20
update control_toolbox branch

urfeex:
control_toolbox has branched out for Humble. This updates the repos file accordingly.

See https://github.com/ros-controls/control_toolbox/pull/265


urfeex:
Humble-semi-binary-main will be failing until the next sync, as generate_parameter_library was updating the header-install location as part of an opt-in method (See #1234). ros2_control has adopted this scheme already and we don't have generate_parameter_library in our upstream workspace.

urrsk (approved):

PR #1257, opened 2025-01-29
Remove urdf folder

urfeex:
That has been falsely added in 92b1bea47a939a266378f3095c1a87a6176f8537

This basically replaces #1256


SarveshMalladi (approved):

urfeex:
@pschmutz before merging this - could I ask you to confirm that I am not missing anything about this? Do you require that file as you've opened #1256 ?

pschmutz:
Thank you for looking into this and your question.

I was using Jazzy before and the [ur.ros2_control.xacro](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/urdf/ur.ros2_control.xacro) from the ur_robot_driver package. When I downgraded I found that the file is not available in the Humble package. That was the motivation for my Pull Request

I can use the ur.ros2_control.xacro from the Humble description package, no problem.

PR #1265, opened 2025-02-04
Update computeCommand to compute_command

urfeex:
The former has been marked deprecated in control_toolbox.


urfeex:
Waiting for this until the next sync, as it is not worth it breaking binary builds in CI for removing a deprecation warning.

urfeex:
Upstream control_toolbox has been released. That means, we can unblock this.

urrsk:
@urfeex is the 'Coverage Build' expected to fail?

urrsk (approved):

PR #1274, opened 2025-02-14
Migration of integration tests to pytest

URJala:
The tests in robot_driver.py run, and initialisation is done entirely in conftest and test_common

### DIFF
--- a/ur_robot_driver/test/conftest.py
+++ b/ur_robot_driver/test/conftest.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+# Copyright 2025, Universal Robots A/S
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#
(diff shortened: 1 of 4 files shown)


urfeex:
@URJala is this ready to review?

URJala:
Yes, we talked about it maybe not being ideal to re-launch the driver between each test case, which is the only way I was able to make it work. You said you would take a look, and see if you could find anything that would keep the driver between test-cases. Otherwise feedback on the structure of the conftest and robot driver package is also welcome

urfeex:
How would I supposed to be running that? I am not that familiar with pytest / launch_pytest so I might just be stupid, but the standard `colcon test` seems not to work. Looking at launch_pytest I am not even sure whether a CMake integration is possible?

URJala:
I run it with `pytest path/to/testfile.py` (with options -s -v for more output, and --full-trace for a stack trace in case of errors). Yeah, whenever I have been running any tests locally, I launch ursim myself and remove the ursim and wait_dashboard_server from test_common. That has never worked locally for me.

URJala:
Closing this, as pytest doesnt seem to do, what we want it to do.

PR #1275, opened 2025-02-16
Apply renaming of member variables of JTC

christophfroehlich:
See https://github.com/ros-controls/ros2_controllers/pull/1032


urfeex (approved):
Thank you @christophfroehlich

PR #1283, opened 2025-02-25
Cast to double for future variant API

saikishor:
This would be needed in order to have the controllers compiling with the future API of Handles with variants


saikishor:
@urfeex I believe the failing jobs are not related to this change

urfeex:
Thank you for providing this. Yes, the broken CI jobs are mainly due to #1275 and the big regression currently going on in the buildfarm. Nevertheless I would like to hold this back until things have normalized a bit. @saikishor is there an estimate from when the API we are currently use will be removed from rolling?

saikishor:
> Thank you for providing this. Yes, the broken CI jobs are mainly due to #1275 and the big regression currently going on in the buildfarm. Nevertheless I would like to hold this back until things have normalized a bit. @saikishor is there an estimate from when the API we are currently use will be removed from rolling?

@urfeex I'm planning to add support for more types in ros2_control. You can find more info here : https://github.com/ros-controls/ros2_control/pull/2065

As the default expects double, it is important that it is of the same type. Apart from this, nothing should change in your current setup

urfeex (approved):

PR #1287, opened 2025-03-04
Update transformForceTorque to handle wheter it is a cb3 or an e-Series robot

urfeex:
The force torque is returned at the tool flange on e-series robots and at the tcp for CB3, this is now handled correctly, so that all force/torque measurements will be relative to the active TCP


urmahp (approved):

URJala (approved):

URJala:
Tested on robot, and it seems to work.

PR #1289, opened 2025-03-05
Update ros2_control API to get_optional

urfeex:
Update to new `get_optional` API from ros2_control. Let's try to keep the warnings at 0.

### DIFF
--- a/ur_controllers/src/freedrive_mode_controller.cpp
+++ b/ur_controllers/src/freedrive_mode_controller.cpp
@@ -193,15 +193,15 @@ ur_controllers::FreedriveModeController::on_deactivate(const rclcpp_lifecycle::S
 controller_interface::return_type ur_controllers::FreedriveModeController::update(const rclcpp::Time& /*time*/,
                                                                                   const rclcpp::Duration& /*period*/)
 {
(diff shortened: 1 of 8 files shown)


ur_controllers/src/freedrive_mode_controller.cpp
```diff
@@ -193,15 +193,15 @@ ur_controllers::FreedriveModeController::on_deactivate(const rclcpp_lifecycle::S
 controller_interface::return_type ur_controllers::FreedriveModeController::update(const rclcpp::Time& /*time*/,
                                                                                   const rclcpp::Duration& /*period*/)
 {
-  async_state_ = async_success_command_interface_->get().get_value();
+  async_state_ = async_success_command_interface_->get().get_optional().value();
```
urmahp, line 196: Are we sure that this will always contain a value, because otherwise it can throw an exception?

urmahp (changes requested):

ur_controllers/src/freedrive_mode_controller.cpp
urfeex: We could also use the _or option here, I think. I will update.

PR #1313, opened 2025-03-24
Start executing passthrough trajectories earlier than all points are transferred.

urfeex:
When transferring trajectories with many waypoints it can take quite a
while until all points are transferred from the controller to the
hardware interface, especially on lower hardware_interface update rates.

With this change, execution starts as soon as enough trajectory points
are transferred to fill 5 control cycles.

This PR effectively allows reducing the hardware_interface update_rate
when using the passthrough_trajectory_controller significantly. This
will reduce the networking requirements improving the overall user
experience.

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -1361,12 +1361,8 @@ void URPositionHardwareInterface::check_passthrough_trajectory_controller()
     trajectory_times_[point_index_received] = passthrough_trajectory_time_from_start_ - last_time;
     last_time = passthrough_trajectory_time_from_start_;
 
-    // if (!std::isnan(passthrough_trajectory_velocities_[0])) {
     trajectory_joint_velocities_[point_index_received] = passthrough_trajectory_velocities_;
-    //}
-    // if (!std::isnan(passthrough_trajectory_accelerations_[0])) {
     trajectory_joint_accelerations_[point_index_received] = passthrough_trajectory_accelerations_;
-    //}
 
     point_index_received++;
     passthrough_trajectory_transfer_state_ = 1.0;
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -1392,19 +1392,41 @@ void URPositionHardwareInterface::check_passthrough_trajectory_controller()
   // already received down to the hardware.
   if (trajectory_started && point_index_sent <= trajectory_joint_positions_.size() &&
       point_index_sent < point_index_received) {
+    bool error = false;
     /* Write the appropriate type of point depending on the combination of positions, velocities and accelerations. */
     for (size_t i = point_index_sent; i < point_index_received; i++) {
-      if (!has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
-        ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], urcl::vector6d_t{ 0, 0, 0, 0, 0, 0 },
-                                               trajectory_times_[i]);
-      } else if (has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
-        ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], trajectory_joint_velocities_[i],
-                                               trajectory_times_[i]);
-      } else if (has_velocities(trajectory_joint_velocities_) && has_accelerations(trajectory_joint_accelerations_)) {
-        ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], trajectory_joint_velocities_[i],
-                                               trajectory_joint_accelerations_[i], trajectory_times_[i]);
+      if (is_valid_joint_information(trajectory_joint_positions_)) {
+        if (!is_valid_joint_information(trajectory_joint_velocities_) &&
+            !is_valid_joint_information(trajectory_joint_accelerations_)) {
+          ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], urcl::vector6d_t{ 0, 0, 0, 0, 0, 0 },
+                                                 trajectory_times_[i]);
+        } else if (is_valid_joint_information(trajectory_joint_velocities_) &&
+                   !is_valid_joint_information(trajectory_joint_accelerations_)) {
+          ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], trajectory_joint_velocities_[i],
+                                                 trajectory_times_[i]);
+        } else if (!is_valid_joint_information(trajectory_joint_velocities_) &&
+                   is_valid_joint_information(trajectory_joint_accelerations_)) {
+          RCLCPP_ERROR(get_logger(), "Accelerations but no velocities given. If you want to specify accelerations with "
+                                     "a 0 velocity, please do that explicitly.");
+          error = true;
+          break;
+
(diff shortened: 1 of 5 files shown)


ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
 
-    if (!std::isnan(passthrough_trajectory_velocities_[0])) {
-      trajectory_joint_velocities_.push_back(passthrough_trajectory_velocities_);
-    }
-    if (!std::isnan(passthrough_trajectory_accelerations_[0])) {
-      trajectory_joint_accelerations_.push_back(passthrough_trajectory_accelerations_);
-    }
+    // if (!std::isnan(passthrough_trajectory_velocities_[0])) {
+    trajectory_joint_velocities_[point_index_received] = passthrough_trajectory_velocities_;
+    //}
```
urmahp, line 1353:
```suggestion
```
Remove

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
+    trajectory_times_[point_index_received] = passthrough_trajectory_time_from_start_ - last_time;
     last_time = passthrough_trajectory_time_from_start_;
 
-    if (!std::isnan(passthrough_trajectory_velocities_[0])) {
-      trajectory_joint_velocities_.push_back(passthrough_trajectory_velocities_);
-    }
-    if (!std::isnan(passthrough_trajectory_accelerations_[0])) {
-      trajectory_joint_accelerations_.push_back(passthrough_trajectory_accelerations_);
-    }
+    // if (!std::isnan(passthrough_trajectory_velocities_[0])) {
```
urmahp, line 1351:
```suggestion
```
Remove

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
+    trajectory_joint_velocities_[point_index_received] = passthrough_trajectory_velocities_;
+    //}
+    // if (!std::isnan(passthrough_trajectory_accelerations_[0])) {
```
urmahp, line 1354:
```suggestion
```

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
+    trajectory_joint_accelerations_[point_index_received] = passthrough_trajectory_accelerations_;
+    //}
```
urmahp, line 1356:
```suggestion
```

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
+
+    // Once we received enough points so we can move for at least 5 cycles, we start executing
+    if ((passthrough_trajectory_time_from_start_ > 5.0 / static_cast<double>(info_.rw_rate) ||
+         point_index_received == passthrough_trajectory_size_ - 1) &&
+        !trajectory_started) {
+      ur_driver_->writeTrajectoryControlMessage(urcl::control::TrajectoryControlMessage::TRAJECTORY_START,
+                                                trajectory_joint_positions_.size());
+      trajectory_started = true;
+    }
     /* When all points have been read, write them to the physical robot controller.*/
```
urmahp, line 1369: This comment is no longer applicable.

urmahp (changes requested):
Overall looks good, added some comments.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -93,9 +94,10 @@ URPositionHardwareInterface::on_init(const hardware_interface::HardwareInfo& sys
   freedrive_mode_abort_ = 0.0;
   passthrough_trajectory_transfer_state_ = 0.0;
   passthrough_trajectory_abort_ = 0.0;
-  trajectory_joint_positions_.clear();
-  trajectory_joint_velocities_.clear();
-  trajectory_joint_accelerations_.clear();
+  passthrough_trajectory_size_ = 0.0;
+  trajectory_joint_positions_.reserve(32768);
```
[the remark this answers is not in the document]
urfeex: Since we use [`socket_read_binary_integer`](https://www.universal-robots.com/manuals/EN/HTML/SW5_19/Content/prod-scriptmanual/G5/socket_read_binary_integer.htm) to communicate that is the maximum amount of waypoints that can be encoded with that mechanism. Probably, less would be sufficient, but then any number less than that would seem arbitrary, hence I went with that.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,88 @@ void URPositionHardwareInterface::stop_force_mode()
...
-      }
-    } else if (has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
-      for (size_t i = 0; i < trajectory_joint_positions_.size(); i++) {
+      } else if (has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
         ur_driver_->writeTrajectorySplinePoint(trajectory_joint_positions_[i], trajectory_joint_velocities_[i],
                                                trajectory_times_[i]);
-      }
-    } else if (!has_velocities(trajectory_joint_velocities_) && has_accelerations(trajectory_joint_accelerations_)) {
-      for (size_t i = 0; i < trajectory_joint_positions_.size(); i++) {
+      } else if (!has_velocities(trajectory_joint_velocities_) && has_accelerations(trajectory_joint_accelerations_)) {
```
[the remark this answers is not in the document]
urfeex: I absolutely agree. I must have missed that in an earlier review.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1309,59 +1319,80 @@ void URPositionHardwareInterface::stop_force_mode()
...
+
+  // We basically get a setpoint from the controller in each cycle. We send all the points that we
+  // already received down to the hardware.
+  if (trajectory_started && point_index_sent <= trajectory_joint_positions_.size() &&
+      point_index_sent < point_index_received) {
     /* Write the appropriate type of point depending on the combination of positions, velocities and accelerations. */
-    if (!has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
-      for (size_t i = 0; i < trajectory_joint_positions_.size(); i++) {
+    for (size_t i = point_index_sent; i < point_index_received; i++) {
+      if (!has_velocities(trajectory_joint_velocities_) && !has_accelerations(trajectory_joint_accelerations_)) {
```
urmahp, line 1384: Are we sure that positions is always going to be part of the points received from the forward controller? 
If we can end up in a situation where positions is not part of the received points, then I think we should handle it more cleanly with a separate if statement and Logging an error.

urmahp (changes requested):
Overall it looks good. Have small suggestions which is not really related to this PR, but is part of the touched code, so I stumbled across them.

ur_robot_driver/src/hardware_interface.cpp
urfeex: In the current implementation that we have with the passthrough trajectory controller, that is given. But I do see the point that a better handling would be to catch that here. I would actually advocate for the same thing regarding the acceleration but no velocity. If the user wants to specify accelerations combined with 0 velocities that should be done explicitly.

I'll add catching and raising an error.

urmahp (approved):

PR #1318, opened 2025-03-26
Support PolyScopeX robots

urfeex:
Since PolyScopeX robots don't have a dashboard server, we'll have to make sure to

a) not start the dashboard client when interacting with a PolyScopeX robot
b) replace the robot_state_helper's functions to the dashboard server with primary_client API calls.

This requires https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/296 to work.

### DIFF
--- a/ur_robot_driver/src/robot_state_helper.cpp
+++ b/ur_robot_driver/src/robot_state_helper.cpp
@@ -79,7 +79,7 @@ RobotStateHelper::RobotStateHelper(const rclcpp::Node::SharedPtr& node)
   if (robot_version->major > 5) {
     RCLCPP_WARN(rclcpp::get_logger("robot_state_helper"), "Running on a PolyScopeX robot. The dashboard server is not "
                                                           "available, therefore the robot_state_helper cannot start "
-                                                          "PolyScope programs.");
+                                                          "PolyScope programs and restart the safety.");
   } else {
     // Service to restart safety
     restart_safety_srv_ = node_->create_client<std_srvs::srv::Trigger>(
--- a/ur_robot_driver/src/robot_state_helper.cpp
+++ b/ur_robot_driver/src/robot_state_helper.cpp
@@ -361,12 +361,12 @@ void RobotStateHelper::setModeExecute(const std::shared_ptr<RobotStateHelper::Se
       if (headless_mode_) {
         result_->success = safeDashboardTrigger(this->resend_robot_program_srv_);
       } else {
-        // The dashboard denies playing immediately after switching the mode to RUNNING
-        sleep(1);
         if (play_program_srv_ == nullptr) {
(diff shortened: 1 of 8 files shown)


ur_robot_driver/src/robot_state_helper.cpp
```diff
@@ -346,7 +363,12 @@ void RobotStateHelper::setModeExecute(const std::shared_ptr<RobotStateHelper::Se
       } else {
         // The dashboard denies playing immediately after switching the mode to RUNNING
         sleep(1);
```
urmahp, line 365: This should be moved to the else statement, both the comment and the sleep.

ur_robot_driver/src/robot_state_helper.cpp
```diff
@@ -66,28 +68,27 @@ RobotStateHelper::RobotStateHelper(const rclcpp::Node::SharedPtr& node)
   node->declare_parameter("headless_mode", false);
   headless_mode_ = node->get_parameter("headless_mode").as_bool();
 
-  // Service to unlock protective stop
```
urmahp, line 69: Maybe I am missing something, but to me it seems like all the services disappeared, only the services that are dashboard server specific is still kept.

ur_robot_driver/src/robot_state_helper.cpp
```diff
@@ -66,28 +68,27 @@ RobotStateHelper::RobotStateHelper(const rclcpp::Node::SharedPtr& node)
...
+
+  primary_client_ = std::make_shared<urcl::primary_interface::PrimaryClient>(robot_ip_, notifier_);
+
+  primary_client_->start(0, std::chrono::seconds(10));
+  auto robot_version = primary_client_->getRobotVersion();
+
+  if (robot_version->major > 5) {
+    RCLCPP_WARN(rclcpp::get_logger("robot_state_helper"), "Running on a PolyScopeX robot. The dashboard server is not "
+                                                          "available, therefore the robot_state_helper cannot start "
+                                                          "PolyScope programs.");
```
urmahp, line 82:
```suggestion
                                                          "PolyScope programs and restart the safety.");
```

urmahp (changes requested):

ur_robot_driver/src/robot_state_helper.cpp
urfeex: Yes, that is on purpose. Theses were the service-clients to the dashboard server. Functionality of the removed ones got replaced with direct primary client calls.

urmahp: Arh I should have just looked properly in the code, I just assumed they where servers and not clients....

urmahp (approved):

PR #1341, opened 2025-04-11
Hardware interface for sending motion primitives to the robot via the Instruction Executor instead of trajectories

mathias31415:
This PR extends the `ur_robot_driver` package with a hardware interface for sending motion primitives to the robot via the Instruction Executor instead of traditional trajectory execution.

To reuse the state interface functionalities of the original hardware interface in the new `motion_primitive_ur_driver`, these functionalities have been refactored into a shared helper file: `stateinterface_helper.hpp`.

The implementation uses the [`motion_primitives_forward_controller`](https://github.com/StoglRobotics-forks/ros2_controllers/tree/motion_primitive_forward_controller/motion_primitives_forward_controller), which is planned to be integrated into the [`ros2_controllers`](https://github.com/ros-controls/ros2_controllers) repository.

Motion primitives are received by the controller using the `MotionPrimitive.msg` from the [`industrial_robot_motion_interfaces`](https://github.com/StoglRobotics-forks/industrial_robot_motion_interfaces/tree/helper-types) package. A modified version of this message is used here, which includes **additional helper types**:

- `STOP_MOTION`: Interrupts current robot motion and clears the motion queue.
- `MOTION_SEQUENCE_START` and `MOTION_SEQUENCE_END`: Define a sequence block—primitives between these markers are grouped and executed as a single, blended motion sequence. This enables smooth transitions between individual primitives, unlike executing them one by one.

This implementation also depends on a [pending Pull Request](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/303) in the `Universal_Robots_Client_Library`. This PR extends the Instruction Executor with support for `movec` commands, which is required for executing circular Cartesian motions within this driver.

I'd appreciate any feedback or suggestions – thanks in advance!


urfeex:
Thank you for your contribution! I'll have a look at this hopefully soon.

urfeex:
Just a high-level question while I go through this implementation: Why did you decide to add that functionality to another hardware interface rather than integrating it into the existing one?

Integrating it would allow to switch between modes during runtime, which is especially useful e.g. when users want to use freedrive mode in between.

We already have multiple modes in our current hardware interface calling different methods from the `write()` function depending on the operation mode.

mathias31415:
There's no specific reason for that, it's simply how I started implementing it. But you're probably right that it would be more elegant to integrate it into the existing hardware interface.

mathias31415:
@urfeex  I've now integrated the `motion_primitives_ur_driver` into the `hardware_interface`, which allows switching modes during operation. I'm happy to receive any feedback on this!

During the integration, I encountered the following issue:

Both the `hardware_interface` implementation and the `InstructionExecutor` (used to execute motion primitives) depend on a callback that is triggered when a trajectory is completed. In the current setup, the UR driver's callback function is overwritten, which means only one of these callbacks can be active at any given time.

To solve this, I register the `InstructionExecutor`'s callback when motion primitives mode is activated, and restore the `hardware_interface`'s callback when the mode is deactivated. To enable this mechanism, I added a method `registerTrajDoneCallback()` to the `InstructionExecutor` in the `ur_client_library`:

```cpp
void urcl::InstructionExecutor::registerTrajDoneCallback()
{
  driver_->registerTrajectoryDoneCallback(
      std::bind(&InstructionExecutor::trajDoneCallback, this, std::placeholders::_1));
}
```
Let me know if you have any suggestions for a cleaner solution.

ur_robot_driver/examples/send_joint_positions.py
urfeex, line 1: Is this example related to this PR?

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1464,6 +1607,413 @@ bool URPositionHardwareInterface::is_valid_joint_information(std::vector<std::ar
...
+      }
+      case static_cast<uint8_t>(MoprimMotionHelperType::RESET_STOP):
+      {
+        std::lock_guard<std::mutex> guard(moprim_stop_mutex_);
+        if (!new_moprim_reset_available_) {
+          new_moprim_reset_available_ = true;
+          resetMoprimCmdInterfaces();
+        }
+        break;
+      }
```
urfeex, line 1637: Instead of setting a flag and then handling the cancel event in a separate thread, you could instead add a `STOPPING` state to `motion_primitives_forward_controller::ExecutionState`. Then, you could directly run 

```c++
driver_->writeTrajectoryControlMessage(urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL, -1,
                                           RobotReceiveTimeout::millisec(2000));
```
, set the state to `STOPPING` and not wait for the result.

The result callback could then set an atomic flag that is being picked up e.g. in the `read` method.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1464,6 +1607,413 @@ bool URPositionHardwareInterface::is_valid_joint_information(std::vector<std::ar
...
+      default:
+      {
+        std::lock_guard<std::mutex> guard(moprim_cmd_mutex_);
+        if (!new_moprim_cmd_available_) {
+          // Copy command to thread-safe buffer
+          pending_moprim_cmd_ = hw_moprim_commands_;
+          new_moprim_cmd_available_ = true;
+          resetMoprimCmdInterfaces();
+        }
```
urfeex, line 1646: To avoid this mutex, you could use a `realtime_tools::LockFreeQueue` for pushing the commands to the execution thread.

urfeex (changes requested):

urfeex:
@mathias31415 regarding [b4f4dec](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341/commits/b4f4dec33cb6eaa1d87e6c2c899860fb5e5be91c), please let's keep that back for a follow-up PR.

ur_robot_driver/src/hardware_interface.cpp
mathias31415: I added the `STOPPING` mode and removed the separate thread.  
However, I think I need to wrap `current_moprim_execution_status_` in a `RealtimeThreadSafeBox`, since I'm writing to it both in `read()` (via `handleMoprimCommands()`), in `asyncMoprimCmdThread` (via `processMoprimMotionCmd()`), and now also in `trajectory_done_callback()`, while I'm reading from it in `write()`.

The same applies to `ready_for_new_moprim_`.

What do you think about this?

ur_robot_driver/src/hardware_interface.cpp
mathias31415: I replaced the moprim_cmd_mutex_ with realtime_tools::LockFreeSPSCQueue<std::array<double, 25>, 1> and also changed the command and state interfaces from std::vector to std::array<double, 25>.
@urfeex does that look good to you?

ur_robot_driver/src/hardware_interface.cpp
urfeex: Both, `current_moprim_execution_status_` and `ready_for_new_moprim_` can be wrapped into an `std::atomic<>` container. `ready_for_new_moprim_` already is an atomic bool.

ur_robot_driver/src/hardware_interface.cpp
urfeex: Couldn't the `new_moprim_cmd_available_` member be replaced by checking whether the queue is empty in `asyncMoprimCmdThread`? We can always fill the queue in `handleMoprimCommands` and pop from the queue whenever there is data available. From my understanding this would exactly fit the usecase mentioned in the queue's [docstring](https://github.com/ros-controls/realtime_tools/blob/87ad70c542cae1ff18d1e596e6bbcf32071f82f9/realtime_tools/include/realtime_tools/lock_free_queue.hpp#L242).

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -1464,6 +1599,351 @@ bool URPositionHardwareInterface::is_valid_joint_information(std::vector<std::ar
...
+          if (std::isnan(command[i])) {
+            RCLCPP_ERROR(rclcpp::get_logger("URPositionHardwareInterface"), "Invalid motion command: pose contains NaN "
+                                                                            "values");
+            current_moprim_execution_status_ = MoprimExecutionState::ERROR;
+            return;
+          }
+        }
+
+        // 0: Unconstrained mode, 1: Fixed mode
+        // (https://tools.pages.cba.mit.edu/Universal_Robotics_UR10_Robot_Arm/scriptManual-3.5.4.pdf)
```
urfeex, line 1832:
```suggestion
        // (https://www.universal-robots.com/manuals/EN/HTML/SW5_22/Content/prod-scriptmanual/all_scripts/movec_pose_via_pose_toa1.htm)
```

ur_robot_driver/launch/ur_control.launch.py
```diff
@@ -28,6 +28,8 @@
 
 #
 # Author: Denis Stogl
+#
+# Author modifications: Mathias Fuhrer
```
urfeex, line 32: I do understand that you would like to be visible for the changes that you made. As you can see, it's not really common to state every author that makes changes to a file. Though Denis is listed as the only author, *every* line in this file has been touched by someone in the meantime.

I'm not saying you shouldn't do this, I'll leave that up to you, I just wanted to set it a bit into perspective. If you decide to leave it like that, I think it would be a good idea to specify what kind of modifications you made in that file.

ur_robot_driver/src/hardware_interface.cpp
mathias31415: I changed it as suggested.

urfeex (approved):
Looks good to me. A few nitpicks left, but I'll take them to a follow-up.

Just as a note for now:

- Remove logging from perform_switch
- Add moprim interface to compatibility matrix
- Fix end-of-file in controller-yaml

Issue #1349, opened 2025-04-27
Real Time Kernel not possible for Ubuntu 22.04 and Ros2 Humble


becketps:
### Affected ROS2 Driver version(s)

Ros2 Humble

### Used ROS distribution.

Humble

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

Build both the ROS driver and UR Client Library from source

### Which robot platform is the driver connected to.

URSim in a virtual machine

### Robot SW / URSim version(s)

Polyscope 5

### How is the ROS driver used.

Through the robot teach pendant using External Control URCap

### Issue details

# Summary
*Short introduction to the issue and how it impact you and why*
It is not possible to follow the mentioned steps here for ubuntu 22.04 with the standard Kernel 6.8.0 and this description https://docs.ros.org/en/ros2_packages/humble/api/ur_robot_driver/doc/installation/real_time.html

## Issue details
*Detailed description help us understand the problem. Code are welcome!*
There is no Kernel available for 6.8.0 and therefore I took the stabel LTS Version 5.15.179 https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git which I only can build with

make -j $(getconf _NPROCESSORS_ONLN) deb-pkg

The result is an Error 
dpkg-buildpackage: Information: Host-Architektur amd64
 dpkg-source -i.git --before-build .
 debian/rules clean
rm -rf debian/files debian/linux-*
make clean
 dpkg-source -i.git -b .
dpkg-source: Information: Quellformat »1.0« wird verwendet
dpkg-source: Warnung: Quellverzeichnis »linux-5.15.179« lautet nicht <Quellpaket>-<Ursprungsversion> »linux-upstream-5.15.179-rt84«
dpkg-source: Warnung: .orig-Verzeichnisname linux-5.15.179.orig ist nicht <Paket>-<Ursprungsversion> (linux-upstream-5.15.179-rt84.orig erwünscht)
dpkg-source: Information: linux-upstream wird unter Benutzung des existierenden linux-upstream_5.15.179-rt84.orig.tar.gz gebaut
dpkg-source: Information: linux-upstream wird in linux-upstream_5.15.179-rt84-1.diff.gz gebaut
dpkg-source: Fehler: Änderung in vmlinux-gdb.py kann nicht dargestellt werden:
[...]

PR #1354, opened 2025-05-05
Make check_starting_point of test_move launch files configurable

urfeex:
This should allow to use them in situations, where the robot is not at the preconfigured pose, but users understand the risks and want to circumvent this.

Also, I updated the documentation around this.


urfeex:
This recently came up in #1349

urmahp (approved):

PR #1355, opened 2025-05-09
Check links using lychee instead of a custom script

urfeex:
Recently, our check_links  workflow was failing very often, apparently GitHub has introduced new rate limits. I played around with [lychee](https://github.com/lycheeverse/lychee-action), a tool specifically made for link checking, which supports caching and even understands local links in Markdown and RST files.

It is significantly faster and allows caching built-in. And it helped me find three dead (local) links in the robot_driver's README.

Note: I expect things to fail with the initial run, as the cache has to be built up first.


.github/workflows/check_links.yml
```diff
@@ -11,9 +11,30 @@ jobs:
...
+        uses: lycheeverse/lychee-action@v2
+        with:
+          fail: true
+          args: >
+            --verbose
+            --no-progress
+            --cache
+            --cache-exclude-status 429
+            --max-cache-age 2d
+            './**/*.md' './**/*.html' './**/*.rst' './**/*.cpp' './**/*.h' './**/*.py'
```
urmahp, line 32: Probably just me that is not fully into the lychee framework, but was is the purpose of this argument, it doesn't seem to be tied to any argument?

urfeex: That's the files that are going to be checked. I'll change it such that that's the last argument

urmahp (approved):

PR #1394, opened 2025-06-10
Fix semi-binary builds for Kilted & Rolling

urfeex:
Since there are still outstanding upstream problems, let's try to fix them for our CI so we don't have to fly blind anymore.


urfeex:
Since moveit_resources was just released, this will probably not be necessary anymore.

urfeex:
Let's add it. We don't know when the next rolling sync will be and until then we will not have a working semi-binary unless we merge this.

urmahp (approved):

PR #1410, opened 2025-06-16
Use new API of PID class

christophfroehlich:
Once https://github.com/ros-controls/ros2_controllers/pull/1759 and https://github.com/ros-controls/control_toolbox/pull/400 is merged.

Fixes https://github.com/ros-controls/control_toolbox/issues/392


urfeex:
@christophfroehlich we have fixed the semi-binary pipeline, so rebasing this as soon as https://github.com/ros-controls/control_toolbox/pull/400 is merged, should succeed :-)

urfeex (approved):
Thanks again for the heads-up

urfeex:
The only test failing is a flaky test, that should be unrelated to this.

PR #1438, opened 2025-06-26
Fix: Ignore first trajectory point in passthrough traj controller if time is 0

urfeex:
Currently, when the first trajectory point has a time of 0, this raises an error on the robot, as the spline segment would have the length 0.

Some planners do add such a point with the current joint positions at the beginning. An easy fix to handle this is to ignore the first point in that case.

Note: Currently, this contains a commit to configure moveit and the driver to use the passthrough controller by default to easily test this. This will be reverted before this is marked as ready.


urfeex:
Nevermind. We handle this in URScript. This popping up was only due to my test setup where I mock 5 of the 6 joints.

PR #1452, opened 2025-07-10
ur_moveit_config: Assure the description is loaded as string

urfeex:
Otherwise the description might get loaded as yaml under certain situations.

As noted in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1103#issuecomment-3053056137

@katallen405 would you like to test that?

### DIFF
--- a/ur_moveit_config/launch/ur_moveit.launch.py
+++ b/ur_moveit_config/launch/ur_moveit.launch.py
@@ -125,7 +125,7 @@ def launch_setup(context, *args, **kwargs):
             " ",
         ]
     )
-    robot_description = {"robot_description": robot_description_content}
+    robot_description = {"robot_description": ParameterValue(robot_description_content, value_type=str)}
 
     # MoveIt Configuration
     robot_description_semantic_content = Command(


urmahp (approved):

urfeex:
> I am able to build the humble branch, just not the patch branch

That seems highly unlikely given the fact that the differences between the two are only in a launch file. But there could be intransparent side effects if you still have the binary packages installed.

> It is missing ros2_control_test_assets (which I do not know where to find)

The best way to find missing dependencies is rosdep. To find a specific package, you can run

```console
$ rosdep resolve ros2_control_test_assets
#apt
ros-humble-ros2-control-test-assets
```

Or to install all direct dependencies:
```console
$ rosdep install --from-paths . --ignore-src -y
```

When building a package from source it is advisable to remove its binary installations:
```console
$ sudo apt remove ros-humble-ur-robot-driver ros-humble-ur-controllers ros-humble-moveit-config ros-humble-ur ros-humble-ur-calibration ros-humble-ur-dashboard-msgs 
```

Issue #1457, opened 2025-07-14
ur_ros2_control_node not found


karanchahal-nv:
### Affected ROS2 Driver version(s)

jazzy

### Used ROS distribution.

Other

### Which combination of platform is the ROS driver running on.

Ubuntu Linux with standard kernel

### How is the UR ROS2 Driver installed.

From binary packets

### Which robot platform is the driver connected to.

UR E-series robot

### Robot SW / URSim version(s)

Polyscope 5.22

### How is the ROS driver used.

Headless without using the teach pendant, Through the robot teach pendant using External Control URCap

### Issue details

# Summary
We are trying to launch the UR drivers on Polyscope 5.22 but using ROS2 Jazzy instead of ROS2 Humble.

We realized that the `ur_ros2_control_node` node inside the `ur_robot_driver` package does not exist anymore.

Could you help us in pointing out why the change was made and what we should do to properly launch the ur driver now ?

Our earlier code looked like this:

```
ur_control_node = Node(
   package='ur_robot_driver',
   executable='ur_ros2_control_node',
   parameters=[
      robot_description,
      update_rate_config_file,
      ParameterFile(initial_joint_controllers, allow_substs=True),
      ParameterFile(robotiq_gripper_controllers, allow_substs=True),
   ],
   output='screen',
   on_exit=Shutdown(),
)
```
## Issue details
*Detailed description help us understand the problem. Code are welcome!*

## Steps to Reproduce
*Make simple example to reproduce the issue. Try to remove dependencies to other hardware and software components, if it is possible.*

## Expected Behavior
*What did you expect and why?*

## Actual Behavior
*What did you observe? If possible please attach relevant information.*

# Workaround Suggestion
*If a workaround has been found, you are welcome to share it.*

### Relevant log output

```shell
[tool_communication.py-1] [INFO] [1752017837.261150633] [tool_communication]: 192.56.1.2
[tool_communication.py-1] [INFO] [1752017837.262513545] [tool_communication]: Remote device is available at '/tmp/ttyUR'
[tool_communication.py-1] [INFO] [1752017837.263205417] [tool_communication]: Starting socat with following command:
[tool_communication.py-1] socat pty,link=/tmp/ttyUR,raw,ignoreeof,waitslave tcp:192.56.1.2:54321
[...]

PR #1458, opened 2025-07-15
Add migration of ros2_control node to migration notes

urfeex:
That was missed when doing the change and came up recently in #1457


URJala (approved):
Looks good to me

urfeex:
Will wait with merging until #1457 is resolved.

PR #1461, opened 2025-07-18
Trajectory until node

URJala:
Separate node for executing trajectories while also monitoring an until condition. Currently works with any motion controller, that uses the FollowJointTrajectory action, and the tool contact controller.


urfeex (commented):
I think, if we could improve the docs a little bit, this would be good to go!

urfeex:
I'll wait with merging this until ur_msgs 2.3.0 is merged into Rolling's rosdistro

PR #1467, opened 2025-07-25
Use hpp headers from geometry2

urfeex:
geometry2 recently changed their headers to use a .hpp instead of .h file ending. This commit adapts includes from geometry2 to that.

This should fix the current rolling regression in the buildfarm, as we build the controllers with treating warnings as errors.


URJala (approved):

urfeex:
This will break current kilted builds, as the change has not been released there, yet. See https://github.com/ros/rosdistro/pull/47041. Will merge once that rosdistro PR is merged.

urfeex:
With the compatibility layer added, this is actually mergable.

PR #1469, opened 2025-07-28
Release/4.1.0

urfeex:
As usual, I will rebase-merge, push tags and bloom-release once this is approved.

This is a minor bump due to the added service in the dashboard client.


urfeex:
@URJala I would like to sneak in #1470 into that release, as we would have to make yet another patch release immediately after.

URJala (approved):

PR #1484, opened 2025-09-04
Launch arguments support bool instead of string types only

henningkayser:
The priorliy used string comparison would not work when passing proper bool types when including the launch file. This fix allows using both string and bool values.

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch.py
+++ b/ur_robot_driver/launch/ur_control.launch.py
@@ -192,11 +192,11 @@ def launch_setup(context):
         "freedrive_mode_controller",
         "tool_contact_controller",
     ]
-    if activate_joint_controller.perform(context) == "true":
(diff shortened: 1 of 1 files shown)


urfeex (changes requested):
I'm sorry to retract my approval. The method you proposed, seems to fail. When starting from the commandline:

```
ros2 launch ur_robot_driver ur5e.launch.py robot_ip:=192.168.56.101  use_mock_hardare:=False
```

The expression `if use_mock_hardware.perform(context)` seems to evaluate to `True`. That probably is, since `use_mock_hardware` is read as a string from the command line, probably because its default value is `"false"`. However, if I make the default value a boolean `False`, the `IfCondition(use_mock_hardware)` fails with `'bool' object is not iterable`.

I don't know how to resolve this for now.

PR #1551, opened 2025-10-25
Allow using the actual torques as joint efforts

### LINKED ISSUE #1468: Add joint torques instead of current in UR ros2 hardware interface
Currently, the UR hardware interface fills the effort interface with the "actual current" from rtde data package. 
Following [this](https://sdurobotics.gitlab.io/ur_rtde/api/api.html#_CPPv4N7ur_rtde4RTDE12RobotCommand4Type17GET_JOINT_TORQUESE) discussion, it seems there is a method to get the joint torque compensated by friction and gravity using [get_joint_torque](https://sdurobotics.gitlab.io/ur_rtde/api/api.html#_CPPv4N7ur_rtde4RTDE12RobotCommand4Type17GET_JOINT_TORQUESE). Is it possible to use this method inside the ros2 driver?

This would be very useful, as the alternative method to obtain joint torques involves converting the actual motor current using the torque constant and then subtracting the effects of friction and gravity — which also requires developing a friction model.

urfeex:
Add the option to use the joint_torques reported through RTDE as joint efforts as discussed in #1468.

This will not run on robots with software version < 5.23.0 / 10.11.0, hence the draft. The RTDE output for actual joints doesn't exist before. We would need to find a solution for that. Possible solutions are:
- Keep a separate output recipe and leave passing a compatible recipe file when this option is enabled to the user (I don't quite like that)
- read the recipe file in the driver, append the torque field if requested and save it as a temporary file that gets passed to the driver object
- Extend the driver config to accept a string vector, as well. Then, we can easily append the field in the hardware interface

---

> [!NOTE]
> **High Risk**
> Default joint effort semantics change and driver init fails on PolyScope older than 5.23.0 / 10.11.0 unless `use_currents_as_efforts:=true`; downstream nodes that assumed motor currents in `JointState.effort` may misinterpret data.
> 
> **Overview**
> **Joint effort state now defaults to physical torques** instead of motor currents on `/joint_states` and ros2_control effort interfaces. Unless `use_currents_as_efforts` is `true`, the driver reads RTDE `actual_current_as_torque` (PolyScope ≥ 5.23.0 / 10.11.0).
> 
> The hardware interface loads RTDE recipes as vectors, **appends** `actual_current_as_torque` when torque mode is selected, and passes those recipes into `UrDriver` (replacing recipe file paths in driver config). Configure-time checks enforce minimum software versions, warn on CB3, and surface clear errors if the robot rejects the RTDE key. The `read()` loop selects `actual_current` vs `actual_current_as_torque` based on the flag.
> 
> **Configuration surface:** new `use_currents_as_efforts` hardware parameter (default `false`), xacro/URDF wiring, and launch arguments on `ur_control.launch.py` / `ur_rsp.launch.py`. Docs and Makoa migration notes describe the behavior change and how to keep legacy currents.
> 
> **Testing:** new parametrized integration test `integration_test_joint_efforts.py` (registered in CMake) validates both effort sources via magnitude heuristics on URSim.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff2a53a7dcdcc6b0dea2b8cbee21920090cb906e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ur_robot_driver/doc/hardware_interface.rst
+++ b/ur_robot_driver/doc/hardware_interface.rst
@@ -29,13 +29,8 @@ The UR hardware interface supports the following control modes:
 - **Effort state interfaces**: Joint efforts are reported as torques by default
   (``actual_current_as_torque``, requires PolyScope >= 5.23.0 / 10.11.0). Set
   ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
+  See :doc:`../hardware_interface_parameters` for more information.
 
-  .. note::
-
-     The torques reported are gravity-compensated, meaning a torque of 0.0 corresponds to a joint that is
-     not moving and does not experience an external load. This will only be true for a correctly setup
-     payload on the robot. Please see :ref:`the GPIO controller <io_and_status_controller>` for more
-     information on how to set up the payload dynamically.
 - **Force control**: The robot's end-effector is controlled by specifying target forces
   in Cartesian space.
 - **Freedrive mode**: The robot can be moved freely by the user without any active control.
--- a/ur_robot_driver/doc/hardware_interface.rst
+++ b/ur_robot_driver/doc/hardware_interface.rst
@@ -26,6 +26,9 @@ The UR hardware interface supports the following control modes:
 - **Velocity control**: The robot's joints are controlled by specifying target velocities.
 - **Effort control**: The robot's joints are controlled by specifying target efforts (torques).
   (Only available when running PolyScope >= 5.23.0 / 10.10.0)
+- **Effort state interfaces**: Joint efforts are reported as torques by default
+  (``actual_current_as_torque``, requires PolyScope >= 5.23.0 / 10.11.0). Set
+  ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
 - **Force control**: The robot's end-effector is controlled by specifying target forces
   in Cartesian space.
 - **Freedrive mode**: The robot can be moved freely by the user without any active control.
--- a/ur_robot_driver/doc/hardware_interface.rst
+++ b/ur_robot_driver/doc/hardware_interface.rst
@@ -29,6 +29,13 @@ The UR hardware interface supports the following control modes:
 - **Effort state interfaces**: Joint efforts are reported as torques by default
   (``actual_current_as_torque``, requires PolyScope >= 5.23.0 / 10.11.0). Set
   ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
+  
+  .. note::
+  
+     The torques reported are gravity-compensated, meaning a torque of 0.0 corresponds to a joint that is 
+     not moving and does not experience an external load. This will only be true for a correctly setup
+     payload on the robot. Please see :ref:`the GPIO controller <io_and_status_controller>` for more
+     information on how to set up the payload dynamically.
 - **Force control**: The robot's end-effector is controlled by specifying target forces
   in Cartesian space.
 - **Freedrive mode**: The robot can be moved freely by the user without any active control.
--- a/ur_robot_driver/doc/hardware_interface.rst
+++ b/ur_robot_driver/doc/hardware_interface.rst
@@ -29,10 +29,10 @@ The UR hardware interface supports the following control modes:
 - **Effort state interfaces**: Joint efforts are reported as torques by default
   (``actual_current_as_torque``, requires PolyScope >= 5.23.0 / 10.11.0). Set
   ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
-  
+
   .. note::
-  
-     The torques reported are gravity-compensated, meaning a torque of 0.0 corresponds to a joint that is 
+
+     The torques reported are gravity-compensated, meaning a torque of 0.0 corresponds to a joint that is
      not moving and does not experience an external load. This will only be true for a correctly setup
      payload on the robot. Please see :ref:`the GPIO controller <io_and_status_controller>` for more
      information on how to set up the payload dynamically.
--- a/ur_robot_driver/doc/hardware_interface.rst
+++ b/ur_robot_driver/doc/hardware_interface.rst
@@ -25,11 +25,11 @@ The UR hardware interface supports the following control modes:
 - **Position control**: The robot's joints are controlled by specifying target positions.
 - **Velocity control**: The robot's joints are controlled by specifying target velocities.
 - **Effort control**: The robot's joints are controlled by specifying target efforts (torques).
-  (Only available when running PolyScope >= 5.23.0 / 10.10.0)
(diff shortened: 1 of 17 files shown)


ndunkelb-nasa:
Hi @urfeex , is there any intention to get this merged in the near future? As we are thinking about sim2real implementation, it makes it hard to compare between simulation and hardware. We get most of the information we need time synchronized from `/joint_states`, but having the effort be incorrect according to [the sensor_msgs definition](https://github.com/ros2/common_interfaces/blob/19a8e9239d38d94ec9c269ee4b59e3a038ce1f8a/sensor_msgs/msg/JointState.msg#L6) adds an extra layer of annoyance. We can of course strap something together from existing tools, like the RTDE publisher, or building this from source, but just figured I would see the status of this MR bc it is already out there.

urfeex:
This just hasn't been a priority since I drafted out the concept for this. This is basically the case, since there seemingly was no interest in having this available. But I think, I should be able to get this to a mergeable state pretty quickly.

> - read the recipe file in the driver, append the torque field if requested and save it as a temporary file that gets passed to the driver object
> - Extend the driver config to accept a string vector, as well. Then, we can easily append the field in the hardware interface

We can actually make a combination of both. Since reading the recipe file to a string vector is actually a static method, we can read it in the hardware interface, append the field if possible and then setup the `UrDriver` object with that vector.

ndunkelb-nasa:
That would be great! It seems like this change shouldn't negatively affect anything, so it would be awesome to have as an option! Thanks for taking a look!

urfeex:
I think the state that it currently is in should be mergeable. This PR needs some polishing, especially on the documentation side, though.

We will be branching out for lyrical pretty soon (before merging #1808), then we can also make this the default on Rolling from that point on.

ur_robot_driver/doc/hardware_interface_parameters.rst
```diff
@@ -117,6 +117,22 @@ Tool voltage that will be set as soon as the UR-Program on the robot is started.
...
+------------------------------------------
+
+Selects which RTDE values are exported on the joint effort state interfaces.
+
+* When set to ``false`` (default), the driver reports the joint torques from the RTDE field
+  ``actual_current_as_torque``. This requires PolyScope >= 5.23.0 / 10.11.0. On older software
+  versions, configuring the hardware interface will fail.
+* When set to ``true``, the driver keeps the previous behavior and reports motor currents from
+  ``actual_current`` as efforts.
+* When set to ``true`` on robot software versions < 5.23.0 / 10.11.0, the driver will report
```
urfeex:
```suggestion
* When set to ``false`` on robot software versions < 5.23.0 / 10.11.0, the driver will report
```

ur_robot_driver/doc/hardware_interface.rst
```diff
@@ -26,6 +26,9 @@ The UR hardware interface supports the following control modes:
 - **Velocity control**: The robot's joints are controlled by specifying target velocities.
 - **Effort control**: The robot's joints are controlled by specifying target efforts (torques).
   (Only available when running PolyScope >= 5.23.0 / 10.10.0)
+- **Effort state interfaces**: Joint efforts are reported as torques by default
+  (``actual_current_as_torque``, requires PolyScope >= 5.23.0 / 10.11.0). Set
+  ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
```
urfeex, line 31:
```suggestion
  ``use_currents_as_efforts`` to ``true`` to report motor currents instead.
  
  .. note::
  
     The torques reported are gravity-compensated, meaning a torque of 0.0 corresponds to a joint that is 
     not moving and does not experience an external load. This will only be true for a correctly setup
     payload on the robot. Please see :ref:`the GPIO controller <io_and_status_controller>` for more
     information on how to set up the payload dynamically.
```

ur_robot_driver/doc/hardware_interface_parameters.rst
```diff
@@ -117,6 +117,22 @@ Tool voltage that will be set as soon as the UR-Program on the robot is started.
...
+------------------------------------------
+
+Selects which RTDE values are exported on the joint effort state interfaces.
+
+* When set to ``false`` (default), the driver reports the joint torques from the RTDE field
+  ``actual_current_as_torque``. This requires PolyScope >= 5.23.0 / 10.11.0. On older software
+  versions, configuring the hardware interface will fail.
+* When set to ``true``, the driver keeps the previous behavior and reports motor currents from
+  ``actual_current`` as efforts.
+* When set to ``false`` on robot software versions < 5.23.0 / 10.11.0, the driver will report
```
urfeex:
```suggestion
* When set to ``false`` on robot software versions < 5.23.0 / 10.11.0, the driver will
```

urfeex:
@ndunkelb-nasa I've added some documentation around this feature. If I remember correctly, we've discussed the topic of explicitly documenting gravity compensation with regards to torques which is what I've tried to do here, as well.

ndunkelb-nasa:
Thanks for those docs! When my team tested out the feature (maybe 3 weeks ago), just by building from source on jazzy, and making the couple of changes manually from this branch, we actually saw that the torque in the effort field from the `joint_state_broadcaster` was the full torque, which seems to disagree with this documentation. Have things changed since then?

Just as clarification, when we were stationary with no external loads, the joint efforts were reading large non-zero values, and if we sent say 2 Nm to a forward effort controller, it would read that previous non-zero value + 2. We actually prefer it as the full values, but wanted to make sure we understand what to expect in the future.

urfeex:
> ... we actually saw that the torque in the effort field from the `joint_state_broadcaster` was the full torque, which seems to disagree with this documentation. Have things 

Let me re-check. I might have had that wrong in my mind on Friday.

Edit: You're right, I mixed that up with the torque commands. I'll fix the docs.

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -883,6 +897,17 @@ URPositionHardwareInterface::on_configure(const rclcpp_lifecycle::State& previou
...
 
+  if (!use_currents_as_efforts_) {
+    if ((version_info_.major == 5 && version_info_.minor < 23) ||
+        (version_info_.major == 10 && version_info_.minor < 11) || version_info_.major < 5) {
+      RCLCPP_ERROR(get_logger(),
+                   "Driver configured to use actual torques as efforts, which is not supported by this software "
+                   "version %s. Please use version 5.23.0 / 10.11.0 or newer for this feature.",
+                   version_info_.toString().c_str());
+      return hardware_interface::CallbackReturn::ERROR;
+    }
```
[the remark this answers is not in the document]
urfeex: fair point.

urrsk (approved):

PR #1615, opened 2025-12-18
Fix TCP wrench transform for e-Series: apply inverse on flange_to_tcp transform

li-yunwen:
### Summary

Fix the wrench transformation from flange frame to TCP frame.
The current implementation applies the transform `F_flange_TCP` (KDL naming convention), but the correct wrench conversion requires applying the inverse transform.

### Details

- `ft` is provided by RTDE as a wrench at flange oriented in the base frame.
- The wrench is first rotated into the flange frame correctly by `base_to_flange.M.Inverse() * ft`.
- When converting the wrench from flange to TCP, the inverse of the `flange_to_tcp` transform must be used.
- KDL wrench calculation ref: https://www.orocos.org/wiki/Geometric_primitives.html#toc75

### Validation

- Tested on a real UR3e (e-Series) using the Jazzy branch.
- The previous implementation produces incorrect force orientation when TCP has a non-zero orientation w.r.t. the flange.
- `flange_to_tcp.Inverse()` fixes the issue.

### Notes

The same calculation code exists in other branches.
This PR only targets `jazzy`, since that is the only branch tested locally.

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -1219,7 +1219,7 @@ void URPositionHardwareInterface::transformForceTorque()
     ft = base_to_flange.M.Inverse() * ft;
 
     // Transform the wrench to the tcp frame
-    ft = flange_to_tcp * ft;
+    ft = flange_to_tcp.Inverse() * ft;
   } else {  // CB3
     KDL::Vector vec = KDL::Vector(urcl_target_tcp_pose_[3], urcl_target_tcp_pose_[4], urcl_target_tcp_pose_[5]);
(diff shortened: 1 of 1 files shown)


urfeex:
@li-yunwen Thank you again! We verified that this implementation is correct and fixes a bug. Would you mind, re-targeting this to main, though? We add all changes to main and then backport it to the feature-branches semi-automatically.

li-yunwen:
@urfeex Thank you for the update! I reset the forked branch onto main, reapplied the wrench-transformation fix, and force-pushed. The PR now reflects only this change. Let me know if anything else needs adjustment.

PR #1652, opened 2026-01-28
Fix force–torque frame to follow TCP updates (tool0_controller)

CalaW:
After #1635, the robot publishes correct force–torque readings in the TCP frame, but `force_torque_broadcaster` still publishes them in the `tool0` frame.
When the TCP changes, the force–torque vectors remain expressed in the fixed `tool0` frame, so their origin and orientation do not update with the TCP, which is incorrect.
This PR updates `force_torque_broadcaster` to publish force–torque data in the `tool0_controller` frame, which is dynamically updated with TCP changes, ensuring the measurements are always expressed in the correct TCP frame.

### DIFF
--- a/ur_robot_driver/config/ur_controllers.yaml
+++ b/ur_robot_driver/config/ur_controllers.yaml
@@ -79,7 +79,7 @@ force_torque_sensor_broadcaster:
       - torque.x
       - torque.y
       - torque.z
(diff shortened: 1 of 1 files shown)


urfeex (approved):
Correct, sorry for not thinking of this.

PR #1661, opened 2026-02-03
Update forward command controller types

urfeex:
The specialized controllers have been deprecated upstream. Instead, users should use forward_command_controller/ForwardCommandController and specify the control domain through the "interface_name" parameter.

See https://github.com/ros-controls/ros2_controllers/pull/1913 for example.


urrsk (approved):

urfeex:
Actually, I think before merging this, we should branch out for Kilted. Technically, this change could be considered a breaking change if anyone explicitly queried for the controller type of one of the forwarding controllers.

I'll create the `kilted` branch, then we can merge this onto `main` directly and I will follow-up with a PR establishing the tooling around backporting to kilted and updating CI.

PR #1707, opened 2026-03-12
RTDE Package to create ROS2 topics for the RTDE Field Names selected.

srvald:
**Summary**

This PR introduces a new ROS 2 package ur_rtde_client that connects to a Universal Robots controller using RTDE and publishes ROS 2 topics for the selected RTDE field names (configured at launch time).

**How to execute it**

The launch file requires two arguments:

- robot_ip: The IP address of the UR controller.
- output_recipe: A YAML list of RTDE field names to publish. The list of available field names is documented in the official [UR RTDE Documentation](https://docs.universal-robots.com/tutorials/communication-protocol-tutorials/rtde-guide.html).

Example:

`ros2 launch ur_rtde_client rtde_client.launch.xml robot_ip:=172.17.0.3 output_recipe:="["payload", "robot_mode"]"`

This will create the following topics:

`/rtde/payload`
`/rtde/robot_mode`

**Package Structure**

- publisher.*
Loads the YAML mapping (rtde_map.yaml), creates the ROS 2 publishers at runtime, and publishes the fields extracted from the RTDE data package.

- rtde_manager.*
Manages the UR RTDE client lifecycle, performs initialization, and spawns a background thread that continuously fetches RTDE data when it is available.

- converter.*
Converts UR RTDE field values to the corresponding ROS 2 message types.

- rtde_client_node.*
ROS 2 node entrypoint. Reads launch parameters, initializes the publisher + manager, and starts communication.

- rtde_map.yaml
Defines how each RTDE variable is mapped to a ROS 2 topic and message type.

### DIFF
--- a/ur_rtde_client/LICENSE
+++ b/ur_rtde_client/LICENSE
@@ -1,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2026, Sergi Romero
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+   contributors may be used to endorse or promote products derived from
+   this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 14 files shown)


ur_rtde_client/src/rtde_client_node.cpp
```diff
@@ -0,0 +1,99 @@
...
+  // Create publishers based on output_recipe and config
+  publisher_->createPublishersForRecipe(output_recipe_);
+  effective_keys_ = publisher_->effective_keys();
+
+  RCLCPP_INFO(this->get_logger(), "Created %zu publishers from recipe", effective_keys_.size());
+
+  try {
+    // Initialize RTDE communication
+    // The RtdeManager will run data reading in a background thread
+    rtde_manager_ = start_communication(*this, robot_ip_, effective_keys_, *publisher_);
```
urfeex: Instead of storing the publisher in a unique ptr and passing a reference, it might be better to use a shared pointer in the first place.

ur_rtde_client/LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2026, Sergi Romero
```
urfeex:
```suggestion
Copyright (c) 2026, Universal Robots A/S
```

urfeex (changes requested):
Thank you for that PR! I think, in general the structure is really good.

I see two major things that I would like to improve:

1. I don't quite understand the necessity of the rtde_client_node. It creates a publisher and a manager and then passes the publisher to the manager. I think, all the logic requiring the publisher should not be in the manager, but in the client node. From my understanding, at this point the manager would only consist of the RTDEClient and would then be obsolete.
2. We can improve a lot on runtime. In the RTDE processing pipeline we should reduce the following as much as possible:
    - memory allocations
    - unnecessary data copies
    - expensive operations such as `std::map::find()`

Another feature that I would like to see is passing the publishing frequency as a parameter (Defaulting to 500Hz) that will then be used for the RTDE setup.
 
Adding a test that we can publish a full recipe with all possible fields would be great.

urfeex:
This feature has been moved to its own repository. See https://github.com/UniversalRobots/RTDE_ROS2_Publisher

Closing this PR.

PR #1767, opened 2026-04-16
Update urscript interface to use primary client

urfeex:
Use a full PrimaryClient inside the urscript_sender node.

Before, it was using a secondary stream directly, which was a bit
error-prune and not that well tested as the primary client is. Also, with the bare stream we don't have any chance of doing error detection.

I also added detection for local control mode. When attempting to send script code to the primary interface while it is
connected to the read-only socket (e.g. because the robot is in local
control mode), the controller will send a A210 error. If we receive this, reconnect the primary client. This automatic reconnection behavior can be switched off by a parameter. A ROS service to trigger reconnection has been added, as well.

### DIFF
--- a/ur_robot_driver/CMakeLists.txt
+++ b/ur_robot_driver/CMakeLists.txt
@@ -223,6 +223,7 @@ install(PROGRAMS
   examples/force_mode.py
   examples/send_dummy_motion_primitives_ur10e.py
   examples/move_until_example.py
+  examples/urscript_sender_example.py
   DESTINATION lib/${PROJECT_NAME}
 )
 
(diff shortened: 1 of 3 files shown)


urrsk (approved):

brta-nhui:
Thanks for this Felix, it looks like this solves some stability issues that have been irking us. It'd be great to get this backported to humble once merged!

urfeex:
@brta-nhui We're currently doing some more changes in the ur_client_library which will affect this. This is basically, why this isn't merged, yet.

urfeex:
Closing this in favor of #1833

PR #1771, opened 2026-04-20
Verify the robot type on startup

### LINKED ISSUE #842: Avoiding mismatches in robot model between drivers and actual robot
### Context
I am controlling a simulated robot from ROS2 and found there was a significant mismatch (in the order of centimetres) between the cartesian position of the tcp on the simulated teach pendant compared to the output from tf2.

I went down a whole rabbit hole trying to fix the issue, including ensuring that there were no tcp offsets, that I was using the correct tf frames, etc.; however this only caused more confusion.

I am using the docker image of UR Sim and the UR ROS2 driver binaries to control the simulated robot. I ran the following commands to start the simulator and drivers:

```
docker run --rm -it -e ROBOT_MODEL=UR10e universalrobots/ursim_e-series
```

```
ros2 launch ur_robot_driver ur10e.launch.py robot_ip:=172.17.0.2
```

### What I was doing wrong
The issue turned out to be a mismatch in robot model between the simulator and ROS. I was passing `ROBOT_MODEL=UR10e` to ur sim, however this caused it to default to using a UR5, since the parameter should not include the trailing "e". Neither UR Sim nor the ROS2 drivers gave any warning or indication of the mismatch in robot model.

### What could be added to prevent this issue in future
It may be worthwhile to look into adding some checks into the drivers to ensure that the loaded URDF model of the robot actually matches the robot that the drivers are connecting to. This would avoid user error such as mine and could protect from more dangerous situations where a real robot is controlled with completely incorrect kinematics.

I talked to @gavanderhoorn while looking into the issue and he may be able to also give some input with regards to potential improvements to the drivers in this regard.

urfeex:
Check the ur_type parameter given from ROS with the type that the driver is actually connected to. If they don't match, hardware configuration fails.

This fixes #842.

At the current state this is just a PoC. I think, I would like to move the helper functions to the client_library. What do you think @urrsk?

Also, the series deduction seems a bit janky, but to my knowledge the series isn't directly available. And since e.g. a ur5 and a ur5e both report "UR5" (or `1`) as robot type, we need to know which one it is.

## ToDo

- [x] Add migration notes (It is a behavior-changing change that the hw interface won't configure)
- [x] Test on all models


ur_robot_driver/CMakeLists.txt
```diff
@@ -9,6 +9,15 @@ option(
...
 )
 
+# This test spins up a fresh URSim container for every supported robot model
+# (and one model-mismatch case), which takes a long time. It is therefore kept
+# behind its own flag so a normal integration-test run is not slowed down by it.
+option(
+  UR_ROBOT_DRIVER_RUN_ROBOT_MODELS_TEST
+  "Run the integration test that runs the driver against URSim for every supported robot model"
+  OFF
+)
```
urrsk, line 19: I think it is great that this check is behind a flag. Though where do we actually us this flag and thereby enabling this test in the CI?

- [x] Enable the test in CI

urfeex: I enabled the test in release PRs

urfeex: I enabled the test in release PRs

urrsk (approved):
Doing CI than only runs when we do a release, is always exciting see running when we do the release!

PR #1779, opened 2026-05-06
Kilted verify robot model

urfeex:
This is a manual backport of #1771 making the new feature opt-in rather than opt-out. This way, it will not change the behavior of existing systems.


urrsk (approved):

urfeex:
@Mergifyio backport humble jazzy

PR #1807, opened 2026-06-01
Make GPIO controller publishers realtime safe

s4lmon:
We are finding great use of these GPIO publishers but they bog down ros2 control as they run inline on the controller manager update loop (i.e. 200 Hz) which we are not finding achievable at all, as the realtime middleware layer forces reliable publish.

This change wraps the publishers in `try_publish` which is used by most stock `ros2_control` broadcasters (`joint_state_broadcaster` etc.).  If contention is occur we allow a `false` return and hand off to a separate publishing thread instead of blocking and stalling the cycle. 

This fix works well for us after testing. We could also just push for the `io_pub_` and `tool_data_pub` to be implemented in this way.

---

> [!NOTE]
> **Medium Risk**
> Touches robot I/O and command services (set_io, payload, hand_back_control) and changes when topics are available vs inactive; publishing may drop or delay under contention instead of blocking.
> 
> **Overview**
> Makes GPIO controller publishing **non-blocking** in the `ros2_control` update loop so high-rate reliable DDS publish no longer stalls the controller manager (e.g. at ~200 Hz).
> 
> All five GPIO-related publishers are wrapped in `realtime_tools::RealtimePublisher` and the update path uses **`try_publish`** instead of synchronous `publish`. Robot mode, safety mode, and program-running topics use a **`try_publish_on_change`** helper so cached message state only advances after a successful publish (retries on the next cycle if the realtime lock is contended).
> 
> **Lifecycle / DDS:** Publisher and service registration moves from **`on_activate`** to **`on_configure`**, with full reset in a new **`on_cleanup`** (including all service handles). **`on_deactivate`** no longer tears down DDS objects, avoiding blocking work on the realtime activate/deactivate path. Service handlers gain **`ensureActive`** to fail requests with `success=false` while the controller is inactive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea893eb1fb1296125289b28f74d9c15a5a722594. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ur_controllers/src/gpio_controller.cpp
+++ b/ur_controllers/src/gpio_controller.cpp
@@ -255,8 +255,11 @@ void GPIOController::publishRobotMode()
   auto robot_mode = static_cast<int8_t>(state_interfaces_[StateInterfaces::ROBOT_MODE].get_optional().value_or(0.0));
 
   if (robot_mode_msg_.mode != robot_mode) {
-    robot_mode_msg_.mode = robot_mode;
-    robot_mode_pub_->try_publish(robot_mode_msg_);
+    auto msg = robot_mode_msg_;
+    msg.mode = robot_mode;
+    if (robot_mode_pub_->try_publish(msg)) {
+      robot_mode_msg_.mode = robot_mode;
+    }
   }
 }
 
@@ -265,8 +268,11 @@ void GPIOController::publishSafetyMode()
   auto safety_mode = static_cast<uint8_t>(state_interfaces_[StateInterfaces::SAFETY_MODE].get_optional().value_or(0.0));
 
   if (safety_mode_msg_.mode != safety_mode) {
-    safety_mode_msg_.mode = safety_mode;
-    safety_mode_pub_->try_publish(safety_mode_msg_);
+    auto msg = safety_mode_msg_;
+    msg.mode = safety_mode;
+    if (safety_mode_pub_->try_publish(msg)) {
+      safety_mode_msg_.mode = safety_mode;
+    }
   }
 }
 
@@ -276,8 +282,11 @@ void GPIOController::publishProgramRunning()
       static_cast<uint8_t>(state_interfaces_[StateInterfaces::PROGRAM_RUNNING].get_optional().value_or(0.0));
   bool program_running = program_running_value == 1.0 ? true : false;
   if (program_running_msg_.data != program_running) {
-    program_running_msg_.data = program_running;
-    program_state_pub_->try_publish(program_running_msg_);
+    auto msg = program_running_msg_;
+    msg.data = program_running;
+    if (program_state_pub_->try_publish(msg)) {
+      program_running_msg_.data = program_running;
+    }
   }
 }
 
--- a/ur_controllers/src/gpio_controller.cpp
+++ b/ur_controllers/src/gpio_controller.cpp
@@ -362,7 +362,12 @@ ur_controllers::GPIOController::on_deactivate(const rclcpp_lifecycle::State& /*p
   // joins a worker thread that may be mid-publish, so teardown is in on_cleanup.
   try {
     set_io_srv_.reset();
+    set_analog_output_srv_.reset();
     set_speed_slider_srv_.reset();
+    resend_robot_program_srv_.reset();
+    hand_back_control_srv_.reset();
+    set_payload_srv_.reset();
+    tare_sensor_srv_.reset();
   } catch (...) {
     return LifecycleNodeInterface::CallbackReturn::ERROR;
   }
--- a/ur_controllers/src/gpio_controller.cpp
+++ b/ur_controllers/src/gpio_controller.cpp
@@ -42,6 +42,25 @@
 
 namespace ur_controllers
 {
+namespace
+{
+// Publishes only when `field` changes, and commits the new value into `msg` only on a successful
+// try_publish. If the publish is dropped (lock contention), `msg` is left unchanged so the change is
+// retried on the next cycle rather than being silently lost.
+template <typename MsgT, typename FieldT, typename ValueT>
+void try_publish_on_change(const std::shared_ptr<realtime_tools::RealtimePublisher<MsgT>>& pub, MsgT& msg,
+                           FieldT MsgT::*field, ValueT new_value)
+{
+  if (msg.*field != new_value) {
+    MsgT candidate = msg;
+    candidate.*field = new_value;
+    if (pub->try_publish(candidate)) {
+      msg.*field = new_value;
+    }
+  }
+}
+}  // namespace
+
 controller_interface::CallbackReturn GPIOController::on_init()
 {
   try {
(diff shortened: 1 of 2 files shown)


ur_controllers/src/gpio_controller.cpp
```diff
@@ -42,6 +42,25 @@
...
+void try_publish_on_change(const std::shared_ptr<realtime_tools::RealtimePublisher<MsgT>>& pub, MsgT& msg,
+                           FieldT MsgT::*field, ValueT new_value)
+{
+  if (msg.*field != new_value) {
+    MsgT candidate = msg;
+    candidate.*field = new_value;
+    if (pub->try_publish(candidate)) {
+      msg.*field = new_value;
+    }
+  }
```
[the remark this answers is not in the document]
s4lmon: These are latched topics that represent current state, not an event loop. So this is the correct behaviour.  This helper means sustained requests will always succeed eventually, without blocking the control loop which was the original behaviour.

ur_controllers/src/gpio_controller.cpp
```diff
@@ -336,16 +354,27 @@ ur_controllers::GPIOController::on_activate(const rclcpp_lifecycle::State& /*pre
 
 controller_interface::CallbackReturn
 ur_controllers::GPIOController::on_deactivate(const rclcpp_lifecycle::State& /*previous_state*/)
+{
+  // Publishers survive deactivation: on_deactivate also runs inside the realtime update loop. Destroying publisher
+  // joins a worker thread that may be mid-publish, so teardown is in on_cleanup.
+  try {
+    set_io_srv_.reset();
+    set_speed_slider_srv_.reset();
```
s4lmon: I have leaned this out to not reset the publishing pointers. Ultimately I think this is acceptable because it is more idiomatic (these don't publish except on `update` anyway) and the cost of resetting is high (for the `.join()` on a thread) when they are in use.

ur_controllers/src/gpio_controller.cpp
```diff
@@ -336,16 +354,27 @@ ur_controllers::GPIOController::on_activate(const rclcpp_lifecycle::State& /*pre
 
 controller_interface::CallbackReturn
 ur_controllers::GPIOController::on_deactivate(const rclcpp_lifecycle::State& /*previous_state*/)
+{
+  // Publishers survive deactivation: on_deactivate also runs inside the realtime update loop. Destroying publisher
+  // joins a worker thread that may be mid-publish, so teardown is in on_cleanup.
```
urfeex: I would assume, the same is true for services, isn't it?

urfeex (commented):
Thank you very much for this long overdue change. The gpio controller was one of the first contributions to the ROS driver and I want to disconnect it to multiple pieces in order to separate between I/Os and UR-specific commands.

Nevertheless, taking care about the existing controller is very much appreciated. I just added one small question / concern, otherwise this looks good to me.

ur_controllers/src/gpio_controller.cpp
s4lmon: The thread join issue is only important for the publishers. Services have no worker thread, so tearing them down is cheap. 

In this case, we also shouldn't defer them to `on_cleanup` as this means they would be exposed on an inactive controller and someone could accidentally call them.

s4lmon:
> Thank you very much for this long overdue change. The gpio controller was one of the first contributions to the ROS driver and I want to disconnect it to multiple pieces in order to separate between I/Os and UR-specific commands.
> 
> Nevertheless, taking care about the existing controller is very much appreciated. I just added one small question / concern, otherwise this looks good to me.

Appreciate the feedback. One thing to note: there is still **one** controller overrun of 7 ms on boot, because of the way we spawn the publishers in the realtime control loop (in `on_activate`). Really, we should do this in `on_configure`. If you're happy, I can make this change. This is a better solution but comes with the ergonomic disadvantage where a caller could call a service/try publish to a controller that is not available. The behaviour is safe, however as nothing would happen.

Let me know your thoughts and I can publish this change too.

ur_controllers/src/gpio_controller.cpp
urfeex: But couldn't that service still be stuck in the executor's event loop? If a service call is currently active, that might take a couple of milliseconds in its callback. Would tearing it down here still run immediately?

urfeex:
With your changes, publishers aren't created in `on_activate` anymore, right?

We could do the same with services. With other controllers we did it that way that we fail the service call from its callback if the controller isn't active.

s4lmon:
> With your changes, publishers aren't created in `on_activate` anymore, right?
> 
> We could do the same with services. With other controllers we did it that way that we fail the service call from its callback if the controller isn't active.

Correct. I just made that change, let me know what you think.

ur_controllers/src/gpio_controller.cpp
s4lmon: It would return immediately, even with a callback running.

The service object can still be in use by the executor, but that doesn't stop the teardown. Looking at `rclcpp`,  the shared ownership here decouples the point at which  the controller being done with it from the point where the object gets destroyed. The destruction simply happens later, on the other thread.

ur_controllers/src/gpio_controller.cpp
```diff
@@ -336,16 +357,32 @@ ur_controllers::GPIOController::on_activate(const rclcpp_lifecycle::State& /*pre
...
+ur_controllers::GPIOController::on_cleanup(const rclcpp_lifecycle::State& /*previous_state*/)
 {
   try {
-    // reset publisher
     io_pub_.reset();
     tool_data_pub_.reset();
     robot_mode_pub_.reset();
     safety_mode_pub_.reset();
     program_state_pub_.reset();
-    set_io_srv_.reset();
```
s4lmon, line 347: always goes through `deactivate` so not needed

s4lmon:
@urfeex hi let me know if you get a chance to review. Have been running this for a couple of weeks across a deployment of 5 UR arms (UR3e + UR15) with no issues, including extensive GPIO use.

ur_controllers/src/gpio_controller.cpp
```diff
@@ -38,10 +38,45 @@
...
+
+template <typename ResponseT>
+bool GPIOController::ensureActive(const ResponseT& resp)
+{
+  if (get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Can't accept new requests. Controller is not running.");
+    resp->success = false;
+    return false;
+  }
+  return true;
```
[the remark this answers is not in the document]
urfeex: This is technically correct, but the effect is probably almost neglectable. We can update that in a separate PR for all controllers together.

urfeex (approved):
This is looking good! Thank you for iterating on this!

urfeex:
@Mergifyio backport kilted, humble, jazzy

s4lmon:
Thanks @urfeex, we have run this for many months with great success.

PR #1881, opened 2026-07-13
Add autoconnect parameter to dashboard client

### LINKED ISSUE #1478: Autoconnect parameter
### Feature summary

When the **dashboard_client** is started, it will try to automatically connect to the UR robot. If the robot is not connected yet or off, it will fail and exit. 
Would it be possible to let the user decide if autoconnect? I see that there are 2 services that can connect and disconnect the robot, so it would be possible to connect afterwards

The "problematic" line is this one:
https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/e860145edb80a5cdf735631d7bd2561ed69621c5/ur_robot_driver/src/dashboard_client_ros.cpp#L61

### Possible solution
Add a parameter to autoconnect and let the default **true** so that there is no behavior change:

```cpp
node_->declare_parameter<bool>("autoconnect", true);
bool autoconnect = node_->get_parameter("autoconnect").as_bool();

if(autoconnect){
  connect();
}
```

If implemented, would be great to backport it to Jazzy.

Thank you for your work!

### Tasks
*To complete this issue involves*
- [ ] Implement the feature
- [ ] Make documentation
- [ ] Make Unit test
- [ ] Make example
- [ ] Test on real hardware

aekanman:
## Summary

Addresses [#1478](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1478): when the dashboard client starts, it currently queries the robot version via the primary interface and then connects to the dashboard server (retrying until success). If the robot is offline, the node blocks or fails before any services are available, so `~/connect` cannot be used later.

This adds an `autoconnect` parameter (default `true`, no behavior change) that skips the startup connection when set to `false`. Version discovery and `DashboardClient` construction move into `connect()`, and `~/connect` is registered before any connect attempt. Remaining dashboard services are registered on the first successful connect.

## Behavior

| `autoconnect` | Startup |
|---|---|
| `true` (default) | Same as today: retry until connected or ROS shuts down |
| `false` | Start disconnected; call `~/connect` once the robot is available |

## Changes

- `dashboard_client_ros`: defer primary version + client creation into `connect()`; register `~/connect` early; `initServices()` after first successful connect
- Guard `receive_timeout` updates when `client_` is not yet created
- Launch: `autoconnect` on `ur_dashboard_client.launch.py`
- Docs: document the parameter and `~/connect` behavior
- Tests: cover `autoconnect:=false` in the existing `dashboard_client.py` launch_testing matrix (one of the three URSim variants)

## Test

```bash
# Manual
ros2 launch ur_robot_driver ur_dashboard_client.launch.py robot_ip:=<IP> autoconnect:=false
ros2 service call /dashboard_client/connect std_srvs/srv/Trigger

# Integration (UR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=ON)
colcon test --packages-select ur_robot_driver --ctest-args -R dashboard_client
```

### Test evidence

Manual (e-Series URSim `universalrobots/ursim_e-series` at `192.168.56.101`, dashboard `:29999`):

| Case | Result |
|---|---|
| `autoconnect:=false` against live robot | PASS — `~/connect` available immediately; pre-connect dashboard calls fail cleanly; after `~/connect`, `get_robot_mode` succeeds |
| `autoconnect:=true` (default) against live robot | PASS — auto-connects; `get_robot_mode` works without manual connect |
| Unreachable robot + `autoconnect:=false` | PASS — node stays up; `~/connect` fails after primary retries (existing URCL behavior) |

Integration (`UR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=ON`): full `dashboard_client.py` matrix including `autoconnect:=false` (`latest`/`ur30`) plus existing `true` variants — PASS via launch_testing / `colcon test -R dashboard_client`.

## Related
Closes #1478
Please consider backporting to **jazzy**.

---

> [!NOTE]
> **Medium Risk**
> Changes dashboard client startup and when ROS services become available; default autoconnect=true preserves prior behavior, but callers relying on all services at node start must set autoconnect or call connect explicitly.
> 
> **Overview**
> Adds an **`autoconnect`** parameter (default **`true`**) so the dashboard client can start without blocking on an offline robot. When **`autoconnect`** is **`false`**, the node registers **`~/connect`** immediately and defers primary-interface version discovery, **`DashboardClient`** creation, and all other dashboard services until the first successful connect (via startup retry loop or a manual **`~/connect`** call).
> 
> **`connect()`** is refactored to own the full connect path (including resetting **`client_`** on failure and a lightweight reconnect when **`client_`** already exists). Service registration moves into **`initServices()`**, invoked only after a successful initial connection. **`receive_timeout`** updates are guarded when no client exists yet.
> 
> Launch, docs, and integration tests expose **`autoconnect`** and exercise the disabled path via **`connect_dashboard_client`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acd2f231496c3785876b7727ac2bd6292efdcd9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ur_robot_driver/src/dashboard_client_ros.cpp
```diff
@@ -481,18 +442,63 @@ DashboardClientROS::DashboardClientROS(const rclcpp::Node::SharedPtr& node, cons
...
+    primary_client_.stop();
+    auto dashboard_policy = urcl::DashboardClient::ClientPolicy::G5;
+
+    if (robot_version->major > 5) {
+      if (robot_version->major == 10 && robot_version->minor < 11) {
+        RCLCPP_FATAL(node_->get_logger(),
+                     "The dashboard server for PolyScope X is only available from version 10.11.0 and later. The "
+                     "connected robot has version %s. Exiting now.",
+                     robot_version->toString().c_str());
+        exit(1);
```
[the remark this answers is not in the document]
aekanman: This is intentional and matches current `main` behavior.

Unsupported PolyScope X (`< 10.11`) is a permanent configuration failure for that robot: the dashboard client cannot be used, and retrying `~/connect` would not help. On `main`, that check already calls `exit(1)` during startup, before any services exist.

`autoconnect` only controls **when** the connection is attempted (startup vs later via `~/connect`), not whether an unsupported robot should keep the node alive. Keeping `exit(1)` in `connect()` preserves the same fatal semantics for both `autoconnect:=true` and `autoconnect:=false`.

Changing this to a soft service failure would be a separate behavior change beyond #1478, so I’m leaving the hard exit as-is for now. I can make that change if this is considered a better approach moving forward.

ur_robot_driver/CHANGELOG.rst
```diff
@@ -1,5 +1,6 @@
 7.0.0 (2026-07-09)
 ------------------
+* Add autoconnect parameter to dashboard client (`#1478 <https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1478>`_)
```
urfeex: The CHANGELOG will be filled when making a release.
```suggestion
```

ur_robot_driver/test/dashboard_client_autoconnect.py
```diff
@@ -0,0 +1,104 @@
...
+        }.items(),
+    )
+
+    return LaunchDescription(
+        _declare_launch_arguments()
+        + [ReadyToTest(), dashboard_client, _ursim_action("latest", "ur30")]
+    )
+
+
+class DashboardClientAutoconnectTest(unittest.TestCase):
```
urfeex: Could we potentially merge this with the existing dashboad client test? A completely new test means a new URsim teardown and startup startup in CI.

urfeex (dismissed):
Thank you for contributing this. The changes look quite good in general, but I would like to decrease the test costs for this if possible.

urfeex (changes requested):
See comments above.

ur_robot_driver/test/dashboard_client_autoconnect.py
aekanman: thanks for the review. removed the dedicated `dashboard_client_autoconnect.py` test and folded `autoconnect:=false` into the existing `dashboard_client.py` matrix (covering both cases), so CI still has a single URSim lifecycle for this suite

ur_robot_driver/CHANGELOG.rst
aekanman: thanks, removed

urfeex:
@aekanman could you please correct the formatting issues as pre-commit pointed them out? Otherwise this looks good :-)

urfeex (approved):
Thank you @aekanman for contributing this!

Issue #1658, opened 2026-02-02
Publish the robot's control mode


urfeex:
### Feature summary

We could publish the current control mode as received by the primary interface. It is available in the `RobotModeData` message already.

### Related issues
*Links to related issues internal or external to this repo*

### Tasks
*To complete this issue involves*
- [ ] Implement the feature
- [ ] Make documentation
- [ ] Make Unit test
- [ ] Make example
- [ ] Test on real hardware

urfeex:
@abdela47 Did you find the time to work on this in the meantime? Otherwise I would unassign this again.

ahaffemayer:
Hi @urfeex, I saw your June follow-up on this one. If @abdela47 hasn't had the chance to continue, I'd be glad to take it over. I'm new to this repository but motivated, and your implementation plan above is very clear.

I have time to work on it this coming weekend, so if you're happy to reassign, just let me know and I'll get started.

One thing I wanted to check first: since `ur_msgs` is in `ros-industrial/ur_msgs`, updating `RobotModeDataMsg` with the missing fields would be a PR in that repo as well. Would you like me to open that one first?

urfeex:
@ahaffemayer yes, if you would like to have a go, I think that is fair. Regarding the messages: Usually we do it that way that we open both PRs (one for the msgs and one for the feature in parallel). Then, the message package update gets merged once the feature implementation is approved, such that we know the message is where we would like it to be.
Then, after the msgs package has been released, we merge the feature.

Since updating an existing message, I would be somewhat conservative backporting that to existing distros, which is why the msgs PR should be targeting the `rolling` branch.

@abdela47 I'll reassign this to @ahaffemayer

PR #1919, opened 2026-08-01
[FEAT] Publish robot control mode

ahaffemayer:
Partial implementation of #1658. Step 1-3.5 of the plan. Opening as a draft to check direction before I do the message and finish the broadcaster.

### Content

RobotModeData is read from the primary client in read() and control_mode exposed as gpio/robot_control_mode.

Verified in URSim, 0 at rest, 1 in freedrive, 2 in force mode.

However I have 4 questions: 

- I used `getRobotModeData()` rather than `getRobotMode()`, so the broadcaster can fill the whole `RobotModeDataMsg` from one call. Also robot_mode alone is already on gpio/robot_mode from RTDE. Does that match your intent?
- There's no ControlMode enum in ur_client_library; control_mode_ is a bare uint8_t. Would you prefer one added there alongside RobotMode, or just constants in `ur_msgs/RobotModeDataMsg`?
- `getRobotModeData()` returns null until the first package arrives, so I need a "not received yet" value. Currently a local CONTROL_MODE_UNKNOWN_ = 255. Should that live in RobotModeDataMsg instead so consumers can test for it?
-  getRobotModeData() takes a mutex and a copy on every call, and read() runs at 500 Hz against a source that updates at ~10 Hz if I understood correctly. Should we move it to asyncThread instead? 

### To do
RobotModeDataMsg fields + constants, ur_robot_mode_broadcaster, docs, tests, example, real-hardware check.

### DIFF
--- a/ur_controllers/CMakeLists.txt
+++ b/ur_controllers/CMakeLists.txt
@@ -72,8 +72,8 @@ generate_parameter_library(
   src/gpio_controller_parameters.yaml
 )
 generate_parameter_library(
-  ur_robot_broadcaster_parameters
-  src/ur_robot_broadcaster_parameters.yaml
+  ur_robot_mode_broadcaster_parameters
+  src/ur_robot_mode_broadcaster_parameters.yaml
 )
 
 generate_parameter_library(
@@ -108,7 +108,7 @@ add_library(${PROJECT_NAME} SHARED
   src/speed_scaling_state_broadcaster.cpp
   src/freedrive_mode_controller.cpp
   src/gpio_controller.cpp
-  src/ur_robot_broadcaster.cpp
+  src/ur_robot_mode_broadcaster.cpp
   src/gravity_update_controller.cpp
   src/passthrough_trajectory_controller.cpp
   src/ur_configuration_controller.cpp)
@@ -121,7 +121,7 @@ target_link_libraries(${PROJECT_NAME}
   force_mode_controller_parameters
   friction_model_controller_parameters
   gpio_controller_parameters
-  ur_robot_broadcaster_parameters
+  ur_robot_mode_broadcaster_parameters
   gravity_update_controller_parameters
   speed_scaling_state_broadcaster_parameters
   freedrive_mode_controller_parameters
--- a/ur_controllers/CMakeLists.txt
+++ b/ur_controllers/CMakeLists.txt
@@ -71,6 +71,10 @@ generate_parameter_library(
   gpio_controller_parameters
   src/gpio_controller_parameters.yaml
 )
+generate_parameter_library(
+  ur_robot_broadcaster_parameters
+  src/ur_robot_broadcaster_parameters.yaml
+)
 
 generate_parameter_library(
   gravity_update_controller_parameters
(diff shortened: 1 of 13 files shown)


ahaffemayer:
I've put my own name on the copyright line since I'm contributing personally; happy to change it if you'd prefer a different attribution.

urfeex:
This is definitively going in the right direction. To answer your questions:

> I used getRobotModeData() rather than getRobotMode(), so the broadcaster can fill the whole RobotModeDataMsg from one call. Also robot_mode alone is already on gpio/robot_mode from RTDE. Does that match your intent?

Yes, absolutely. Sorry that I was imprecise in my task description.

> There's no ControlMode enum in ur_client_library; control_mode_ is a bare uint8_t. Would you prefer one added there alongside RobotMode, or just constants in ur_msgs/RobotModeDataMsg?

We can do both, but they are independent of each other. Let's start with adding the constants in the message for now.

> getRobotModeData() returns null until the first package arrives, so I need a "not received yet" value. Currently a local CONTROL_MODE_UNKNOWN_ = 255. Should that live in RobotModeDataMsg instead so consumers can test for it?

I think it would be good if it lived inside the message, yes.

> getRobotModeData() takes a mutex and a copy on every call, and read() runs at 500 Hz against a source that updates at ~10 Hz if I understood correctly. Should we move it to asyncThread instead?

This is definitively a valid concern. This will lead you with the issue of thread synchronization, though. One idea would be to keep the non-double buffer inside an atomic variable, so inside `write()` we can reliably read from that buffer and put data into the double buffer used in the handle.
Even better than using an async thread, you could [register a consumer against the primary client](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/3ca5f04ee8c042c31099960039127c1901d7debe/include/ur_client_library/primary/primary_client.h#L77) that updates the atomic buffer directly (e.g. by passing a simple lambda).

On the same thing: When we add the controller to the standard set we can have it running as an async controller with it's custom rate. Then, we can also remove the `publish_rate` parameter. Speaking about parameters: I think, you don't use (and don't need) the tf_prefix parameter, right? Thus, we could potentially omit parameters for this controller completely.

ahaffemayer:
Hello,

Sorry for the delay, work has been hectic those last weeks.

I have a few questions before wrapping up everything>

- For the control mode constants, we have: 
``` cpp
  uint8 CONTROL_MODE_POSITION=0
  uint8 CONTROL_MODE_TEACH=1
  uint8 CONTROL_MODE_FORCE=2
  uint8 CONTROL_MODE_UNKNOWN=255
```

I verified 0, 1 and 2 against URSim 5.25.2, but I didn't find CONTROL_MODE_TORQUE. Do you know if it has a value already? 


- For now the message is only partly populated, since the broadcaster only fills `control_mode` and the other fields publish 0 / false. Those would require twelve more state interface, should they go into the gpio group or a dedicated one? 


For the realtime access, I went with your second suggestion rather than the async thread, because the lambda on the primary client would need an URCL release as well, since from my reading, `PrimaryConsumer` only exposes
  `setErrorCodeMessageCallback`, `setKeyMessageCallback` and  `setRuntimeExceptionCallback`, there's no setter for `RobotModeData`. Happy to implement it if you want. 

I made a [PR for `ur_msgs` as well.](https://github.com/ros-industrial/ur_msgs/pull/48) 

Still TODO:
Unit testing and the example.

urfeex:
I have reviewed the msgs PR with a couple of change requests. Regarding the control node: Please also see the [primary interface documentation](https://docs.universal-robots.com/tutorials/communication-protocol-tutorials/primary-secondary-guide.html#control-modes). Sorry, I didn't link this earlier.

> For now the message is only partly populated, since the broadcaster only fills control_mode and the other fields publish 0 / false. Those would require twelve more state interface, should they go into the gpio group or a dedicated one?

Yes, I think adding its own GPIO group sounds reasonable. The consumer is perfectly fine.

Thank you for sticking with this :-)

PR #1923, opened 2026-08-04
Report operational status with mock hardware

### LINKED ISSUE #976: [Feature] - Report a valid robot state (robot_mode/safety_state/program_state) when running with mock hardware.
### Feature summary

*Short introduction to the feature and why it should be implemented*

I am currently implementing a diagnostics node. (As you might have guessed from  the other issues). 
At the moment the GPIOController simply doesn't report anything when run with fake/mock hardware as far as I can tell.
It would be nice if it would report:

Robot Mode: Running
Safety Mode: Normal
Program State: true (running)

### Related issues

#974 
#973 

### Implementation

There are two ways to implement this:

A) The GPIOController needs access to the information if it is running with mock hardware and could adapt the status report accordingly

B) It is apparently possible to use: fake_sensor_commands in order to provide arbitrary information on state interfaces. When launched with mock hardware we could make sure that the states listed above are set to the corresponding state interfaces.

Daniiiil1:
## What changed
- initializes mock `robot_mode`, `safety_mode`, and `program_running` state interfaces to RUNNING, NORMAL, and true
- keeps those defaults conditional on `use_mock_hardware`, so real hardware behavior is unchanged
- adds a launch-test assertion against all three latched controller topics

## Why
The GPIO controller already publishes these ordinary state interfaces, but GenericSystem previously initialized them to zero. Mock launches therefore reported disconnected/undefined status and could not support diagnostics tests.

## Impact
This provides deterministic happy-path status for mock hardware without adding mock-specific branches to the production GPIO controller.

## Checks
- repository pre-commit suite passed for both changed files (XML, Black, Flake8, pydocstyle, codespell, copyright, and standard hooks)
- `xmllint --noout`
- Python byte-compilation
- ROS 2 is not installed on the local macOS host, so the launch test is left for hosted ROS CI

Closes #976

---

> [!NOTE]
> **Low Risk**
> Changes are gated on `use_mock_hardware` in xacro and a test-only assertion; real robot hardware paths are untouched.
> 
> **Overview**
> Mock launches previously exposed **disconnected / zero** values on `io_and_status_controller` status topics because `GenericSystem` defaulted GPIO state interfaces to 0. This PR seeds **`robot_mode`**, **`safety_mode`**, and **`program_running`** with happy-path values (**RUNNING**, **NORMAL**, **true**) in `ur.ros2_control.xacro`, only when **`use_mock_hardware`** is enabled.
> 
> A new mock-hardware launch test subscribes to the three latched topics and asserts those values, so diagnostics and CI can rely on stable operational status without mock-specific logic in the GPIO controller.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8afbdb1b2df236534266ea68ab21c0a1e76cf6d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ur_robot_driver/test/test_mock_hardware.py
+++ b/ur_robot_driver/test/test_mock_hardware.py
@@ -37,8 +37,11 @@ import pytest
 import rclpy
 from geometry_msgs.msg import Inertia, Vector3
 from rclpy.node import Node
+from rclpy.qos import DurabilityPolicy, QoSProfile, ReliabilityPolicy
 from control_msgs.action import FollowJointTrajectory
 from controller_manager_msgs.srv import SwitchController
+from std_msgs.msg import Bool
+from ur_dashboard_msgs.msg import RobotMode, SafetyMode
 
 sys.path.append(os.path.dirname(__file__))
 from test_common import (  # noqa: E402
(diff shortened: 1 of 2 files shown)


urfeex (approved):
This is looking good. I assume you used `setdefault()` in the tests in order to explicitly catch the first message published on the topics, which does make sense if you want to target the initial states explicitly.

Issue #1937, opened 2026-08-10
Joint prefix supported in URDF, but not in MoveIt SRDF


mcarmenbg:
### Context

I am working on a simulation with two different robots, one of them a Universal Robots UR3e, and I would like to differentiate the joints of each robot in MoveIt by using a prefix.

I noticed that the URDF provided by `Universal_Robots_ROS2_Description` supports this through the `tf_prefix` parameter, but the SRDF configuration does not appear to provide an equivalent option.

In the URDF macro, as shown in the screenshot, it is possible to provide a prefix through the `tf_prefix` parameter.  Path: `Universal_Robots_ROS2_Description/urdf/ur_macro.xacro`

<img width="395" height="303" alt="Image" src="https://github.com/user-attachments/assets/511ed523-a1d0-499c-9166-f7fff36005b4" />

----------------------------------------------------------------------------------

<img width="363" height="112" alt="Image" src="https://github.com/user-attachments/assets/5392e112-25c3-4002-a90d-4091a152d8f9" />

-------------------------------------------------------------------------------------

However, this does not seem to be possible in the SRDF. Path: `Universal_Robots_ROS2_Driver/ur_moveit_config/srdf/ur_macro.srdf.xacro`

<img width="556" height="692" alt="Image" src="https://github.com/user-attachments/assets/502588d5-15e9-4140-8e6d-54183edbc421" />

### For example:

When I use: `tf_prefix="ur_"`, the URDF generates joint and link names such as:

  - `ur_shoulder_pan_joint`
  - `ur_elbow_joint`
  - `ur_base_link`
  - `ur_tool0`

However, the SRDF still references the unprefixed names:

  - `shoulder_pan_joint`
  - `elbow_joint`
  - `base_link`
  - `tool0`
 

As a result, MoveIt reports that the UR joints and links referenced in the SRDF are not known to the URDF, and the ur_manipulator group cannot be created.

<img width="1077" height="314" alt="Image" src="https://github.com/user-attachments/assets/65056451-9f1f-4725-ad8d-82cbc82b3cb9" />
---------------------------------------------------------------------

<img width="1386" height="174" alt="Image" src="https://github.com/user-attachments/assets/4054a7a8-7305-46fa-b879-fbe964bd5f45" />

----------------------------

### System

- Repository branch: `jazzy`
- Simulator: `Gazebo Sim, version 8.11.0`
- ROS 2 distribution: `Jazzy Jalisco`
[...]

PR #1938, opened 2026-08-10
Add srdf tf prefix

urrsk:
Add tf_prefix support to MoveIt SRDF
The URDF supports a tf_prefix parameter to prefix all link and joint
names, but the MoveIt SRDF hardcoded the unprefixed names. As a result,
using a non-empty tf_prefix caused MoveIt to fail building the
ur_manipulator group because the referenced joints/links did not exist
in the URDF.

Thread a tf_prefix parameter (default empty, fully backward compatible)
through the SRDF macro, its wrapper, and the ur_moveit.launch.py launch
file so the SRDF link/joint references match the generated URDF.

Ref https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1937

### DIFF
--- a/ur_moveit_config/CMakeLists.txt
+++ b/ur_moveit_config/CMakeLists.txt
@@ -3,6 +3,12 @@ project(ur_moveit_config)
 
 find_package(ament_cmake REQUIRED)
 
+if(BUILD_TESTING)
+  find_package(ament_cmake_pytest REQUIRED)
+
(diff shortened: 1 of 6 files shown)


urrsk:
Wrong approach as the moveit_configs_builder already does the job.
See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_moveit_config/doc/migration/jazzy.rst

PR #1939, opened 2026-08-20
Set command interface in tool contact controller

URJala:
Previously, when a tool contact action was accepted and executed, the command interface would stay at TOOL_CONTACT_WAITING_BEGIN, and trigger startToolContact from the hardware interface every write cycle. The command interface is now set to TOOL_CONTACT_EXECUTING when the hardware interface has confirmed it is started. Not sure if this is what was causing flakiness, but it is definitely a bug.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 2ce1326de2307bffdcb841152d81fd6071f7b604. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


urfeex (approved):
That seems reasonable

PR #1940, opened 2026-08-20
Fix/passthrough controller very short trajectories

urfeex:
In order to start trajectories while not all points are transferred, yet, we checked whether the time of the current point is greater than a certain threshold and then start trajectory execution. If the trajectory is shorter than that threshold, nothing would ever be executed.

Also, when the trajectory would finish in between one cycle, our mechanism for resetting the buffers once the trajectory is completely transferred, would never be executed and new trajectories (of the same size) would never get transferred correctly. This PR addresses fixes issues and adds integration tests for these edge cases.


urrsk (approved):
Interesting that this little restructuring fix this!

PR #1950, opened 2026-09-01
[RobotStateHelper] Fix worker thread lifecycle management

### LINKED ISSUE #1943: `RobotStateHelper` detached SetMode workers can run indefinitely, outlive the helper, and corrupt overlapping goals
## Used ROS distribution

Rolling/source. The issue is in driver-owned C++ thread/action state and is not specific to an RMW implementation.

## Platform

Ubuntu Linux (standard, low-latency, or real-time kernels), containers, and other supported source-build platforms share the affected code path. This is a source-level report without a hardware-specific crash reproduction.

## Installation

Build the ROS driver and UR Client Library from source.

## Robot platform

CB3/E-series and real/URSim configurations that instantiate `RobotStateHelper`. No single robot was used for the source-level analysis.

## Robot SW / URSim version(s)

N/A for the ownership defect; the reviewed driver code is affected independently of a specific PolyScope version.

## Driver usage

Both headless mode and teach-pendant/External Control configurations can enter the affected SetMode action paths.

# Summary

`RobotStateHelper::setModeAcceptCallback()` creates a detached thread bound to raw `this` for every accepted SetMode goal. The class has a default destructor and does not retain, stop, or join these threads.

The worker's mode-wait loop has no `rclcpp::ok()` check, stop token, or overall timeout. It only waits for `robot_mode_ == target_mode` or `error_`, but `error_` is initialized to false and is never set anywhere in `ur_robot_driver`. `safeDashboardTrigger()` also calls `future.wait()` without a timeout. A worker can therefore survive executor shutdown indefinitely and later access a destroyed helper.

The goal callback does not reject a goal while `in_action_` is true, and `in_action_` is never reset. Multiple detached workers can overwrite shared `goal_`, `result_`, and `current_goal_handle_`, causing data races and sending a result/feedback to the wrong goal.

## Issue details

The action server accepts every request once robot/safety modes are known. Each accepted goal creates an independent unowned worker, but much of the action state is stored once per helper rather than once per goal.

Although `current_goal_handle_` assignment and some terminal operations take `goal_mutex_`, `goal_` and `result_` are still shared and overwritten by workers. The mutex does not make a multi-goal design correct because a worker can lock and operate on a handle installed by a different worker.

### Indefinite-worker/lifetime path

1. Build the affected commit with AddressSanitizer.
2. Start the driver/URSim and send a valid `~/set_mode` goal.
3. Keep the reported robot mode from reaching the requested target, or make a dashboard service request never complete.
4. Shut down the executor/node and destroy the component containing `RobotStateHelper`.
5. The detached worker remains in the wait loop or future wait and can subsequently access freed helper members.

### Overlapping-goal path

1. Build with ThreadSanitizer.
2. After robot and safety modes are known, send two valid SetMode goals close together.
3. Choose goals that remain active long enough to overlap.
4. Observe concurrent writes to shared action state and possible feedback/result delivery through the wrong `current_goal_handle_`.

## Expected Behavior

SetMode execution has bounded, cancelable lifetime; helper destruction waits for all work to finish; either overlapping goals are rejected or every goal has isolated state and synchronization.

## Actual Behavior

Every accepted goal starts an untracked detached worker. Waits can be unbounded, shutdown is not checked, overlapping goals share mutable state, and helper destruction performs no worker cleanup.

urfeex:
This fixes #1943.

The concept for addressing the issues:
- Do not detach from worker thread, but track it.
- When a new goal is accepted, join the current worker thread before starting a new one.
- When a goal comes in, while there is already one active, reject it.
- When a goal is canceled, make the worker thread stop properly.
- On destruction stop and join the worker thread.
- In the worker thread check also for `rclcpp::ok()`.
- Make the blocking service calls interruptible by checking if stop is requested regularly.

---

> [!NOTE]
> **Medium Risk**
> Changes robot mode transitions, safety recovery, and program start/cancel paths; fixes race/use-after-free risks but alters long-standing SetMode cancel behavior and timing of dashboard calls.
> 
> **Overview**
> Fixes **#1943** by making `SetMode` action execution safe under concurrency, cancellation, and node shutdown.
> 
> **`RobotStateHelper`** is now an **`rclcpp::Node`** (not a helper around a separate node). The implementation is split into **`robot_state_helper_lib`** plus a thin node executable, with new **gmock tests** linked against the library.
> 
> **Worker thread:** detached threads are replaced with a **joined** `worker_thread_`. The destructor sets **`stop_requested_`** and joins. Accepting a new goal joins any prior worker first. **`setModeExecute`** uses **`shouldGoalTerminate()`** (`stop_requested_`, `!rclcpp::ok()`, cancel) at waits and transitions; **`safeDashboardTrigger`** polls service futures and returns **`std::nullopt`** on interrupt. **`setTerminalState`** centralizes succeed / abort / **canceled** outcomes.
> 
> **Action API:** overlapping goals are **rejected** via **`in_action_`** / **`goal_pending_`**; cancel is **accepted** only for the active goal (previously always rejected). Mode/safety feedback is published only when the goal handle is **active**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b3ca15f753fb912906b12064c752ad465c9d7ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ur_robot_driver/src/robot_state_helper.cpp
+++ b/ur_robot_driver/src/robot_state_helper.cpp
@@ -118,7 +118,7 @@ void RobotStateHelper::robotModeCallback(ur_dashboard_msgs::msg::RobotMode::Shar
     RCLCPP_INFO_STREAM(get_logger(), "The robot is currently in mode " << robotModeString(robot_mode_) << ".");
     {
       std::scoped_lock lock(goal_mutex_);
-      if (in_action_) {
+      if (in_action_ && current_goal_handle_->is_active()) {
         feedback_->current_robot_mode =
             static_cast<ur_dashboard_msgs::action::SetMode::Feedback::_current_robot_mode_type>(robot_mode_.load());
         current_goal_handle_->publish_feedback(feedback_);
@@ -134,7 +134,7 @@ void RobotStateHelper::safetyModeCallback(ur_dashboard_msgs::msg::SafetyMode::Sh
     RCLCPP_INFO_STREAM(get_logger(), "The robot is currently in safety mode " << safetyModeString(safety_mode_) << ".");
     {
       std::scoped_lock lock(goal_mutex_);
-      if (in_action_) {
+      if (in_action_ && current_goal_handle_->is_active()) {
         feedback_->current_safety_mode =
             static_cast<ur_dashboard_msgs::action::SetMode::Feedback::_current_safety_mode_type>(safety_mode_.load());
         current_goal_handle_->publish_feedback(feedback_);
--- a/ur_robot_driver/src/robot_state_helper.cpp
+++ b/ur_robot_driver/src/robot_state_helper.cpp
@@ -263,7 +263,7 @@ std::optional<bool> RobotStateHelper::safeDashboardTrigger(rclcpp::Client<std_sr
   auto future = srv->async_send_request(request);
   // Poll in short intervals so a shutdown request is honoured promptly.
   while (future.wait_for(std::chrono::milliseconds(100)) != std::future_status::ready) {
-    if (stop_requested_ || !rclcpp::ok()) {
+    if (shouldGoalTerminate()) {
       srv->remove_pending_request(future);
       return std::nullopt;
     }
@@ -317,7 +317,7 @@ void RobotStateHelper::setModeExecute(const std::shared_ptr<RobotStateHelper::Se
   const auto goal = goal_handle->get_goal();
   this->goal_ = goal;
   urcl::RobotMode target_mode;
-  if (stop_requested_ || !rclcpp::ok()) {
+  if (shouldGoalTerminate()) {
     setTerminalState(false);
     return;
   }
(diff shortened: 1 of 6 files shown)


urrsk (approved):

ur_robot_driver/src/robot_state_helper.cpp
```diff
@@ -42,78 +42,87 @@
...
   primary_client_ = std::make_shared<urcl::primary_interface::PrimaryClient>(robot_ip_, notifier_);
 
   primary_client_->start(0, std::chrono::seconds(10));
   auto robot_version = primary_client_->getRobotVersion();
 
   if (robot_version->major > 5) {
-    RCLCPP_WARN(rclcpp::get_logger("robot_state_helper"), "Running on a PolyScopeX robot. The dashboard server is not "
-                                                          "available, therefore the robot_state_helper cannot start "
-                                                          "PolyScope programs and restart the safety.");
+    RCLCPP_WARN(get_logger(), "Running on a PolyScopeX robot. The dashboard server is not "
```
urrsk, line 80: Would be nice to replace this warning with full support of PolyScopeX and its RobotAPI.
I have made issue #1962 to address this now that we nearly have #1945 in place.

urrsk (approved):

Issue #838, opened 2023-10-10
urscruipt_interface started with wrong IP cannot be stopped normally


fmauch:
While testing around I noticed that when I start the urscript_interface without a connected robot it cannot be stopped easily:

## Steps to reproduce
```
ros2 run ur_robot_driver urscript_interface --ros-args -p robot_ip:=1.2.3.4
```
The output will be
```
ERROR /home/ubuntu/checkout/ur_ros2_rolling/colcon_ws/src/ur_client_library/src/comm/tcp_socket.cpp 129: Failed to
connect to robot on IP 1.2.3.4. Please check that the robot is booted and reachable on 1.2.3.4. Retrying in 10 seconds
```

This will  be repeated every 10 seconds.

I could only stop the process by getting its PID and manually calling `kill -9 $PID`

## Expected behavior
Pressing CTRL-C will terminate the urscript_interface

## Actual behavior
* Pressing CTRL-C doesn't do anything
* Killing the process using SIGTERM
* Killing the process using SIGKILL works

RobertWilbrandt:
Adding a simple handler like
```cpp
rclcpp::on_shutdown([]() { rclcpp::shutdown(); });
```
seems to fix this, but results in
```
terminate called after throwing an instance of 'std::system_error'
  what():  Resource deadlock avoided
```

fmauch:
> `rclcpp::on_shutdown([]() { rclcpp::shutdown(); });`

This seems odd... Honestly, I guess there probably is something to be fixed in the client library. I suspect that our reconnecting mechanism isn't properly interruptible at the moment.

eholum:
Ran into this as well, and FWIW I think [this socket in the constructor](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/src/urscript_interface.cpp#L72) just needs some reasonable values for `max_num_tries ` and a `reconnection_time`, otherwise the constructor sits in [this while loop in the TCP socket](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/src/comm/tcp_socket.cpp#L91) forever. The constructor never completes, so the node never starts spinning.

shin4141:
[...]

Issue #519, opened 2022-10-19
Sending positions to the Hybrid Planner fails, but sending Joint positions works


skaeringur97:
### Description
I followed the [hybrid planning example](https://moveit.picknik.ai/humble/doc/examples/hybrid_planning/hybrid_planning_tutorial.html) and modified it to work with the UR10e robot. It works using the same method as the example uses, but if I try sending positions it always fails in the planning stage. Sending positions works for me on the Panda robot, as well as the Kinova Kortex, but not with the UR10e. I have verified that the position I am sending is valid, by making a [simple MoveIt node](https://gist.github.com/skaeringur97/40956e4c14fdd9acdbbd4d55306ee2ba) that just uses MoveGroupInterface to go to a position. This plans and executes without problems. 

### Your environment
* ROS 2: [Humble] source build
* OS Version: Ubuntu 22.04
* UR10e Binary build (tried source build also)
* Moveit2 Source build

### Steps to reproduce
To reproduce this issue, duplicate the [hybrid planning example](https://moveit.picknik.ai/humble/doc/examples/hybrid_planning/hybrid_planning_tutorial.html) and replace the test/hybrid_planning_demo_node.cpp with [this](https://gist.github.com/skaeringur97/f8165f47531047494b58f4aed394e12e), the test/config/global_planner.yaml with [this](https://gist.github.com/skaeringur97/4c47b3f4895fa21c6b17bdd7e4842d78) and the test/config/local_planner.yaml with [this](https://gist.github.com/skaeringur97/7b366383b0829d91878a56a370ee6b74). The launch file also needs to be changed, so test/launch/hybrid_planning_common.py changes to [this](https://gist.github.com/skaeringur97/5b61c6d6dbfaf95916e701eca4be98d4) 

If you want to see it working with the joint positions, replace these lines:
```
auto const target_pose = []{
        geometry_msgs::msg::PoseStamped msg;
        msg.header.frame_id = "base_link";
        msg.pose.position.x = -0.226126;
        msg.pose.position.y = 0.279057;
        msg.pose.position.z = 1.43127;
        msg.pose.orientation.x = -0.707108;
        msg.pose.orientation.y = 0.000570927;
        msg.pose.orientation.z = 0.000571454;
        msg.pose.orientation.w = 0.707105;
        return msg;
    }();

    std::vector<double> tolerance_pose(3, 0.01);
    std::vector<double> tolerance_angle(3, 0.01);

[...]

PR #1951, opened 2026-09-02
Fix interruptible URScript interface startup

shin4141:
## Summary

- run `PrimaryClient::start(10, 10s)` in a joinable startup worker
- on ROS shutdown during startup, call `PrimaryClient::stop()` and join the worker
- create the existing topic and action endpoints only after startup succeeds
- report genuine startup exceptions as an orderly nonzero process exit
- add a loopback-only process regression test for SIGINT during the retry wait

This addresses #838 without changing retry conditions, URCL, topic/action contracts, controllers, motion, or safety behavior.

## Difference from URCL #519

UniversalRobots/Universal_Robots_Client_Library#519 made the lower-level reconnect path interruptible. This driver change handles the remaining ROS 2 lifecycle gap: `urscript_interface` blocks in `PrimaryClient::start()` before the executor begins, so shutdown must explicitly stop and join that startup operation and must catch startup exhaustion in `main()`.

## Reproduction

```bash
ros2 run ur_robot_driver urscript_interface --ros-args -p robot_ip:=127.0.0.1
# while it is waiting to retry, press Ctrl-C
```

The added launch test reserves loopback port 30001 without listening, waits for the 10-second retry path, sends SIGINT, and requires exit code 0 within 5 seconds.

## Tests

- Format workflow: PASS — https://github.com/shin4141/Universal_Robots_ROS2_Driver/actions/runs/33580143252
- local syntax-only C++ compile, Python compile, clang-format dry-run, and `git diff --check`: PASS
- ROS Rolling driver-local validation: PASS
  - environment: Lima 2.2.0 VM, Ubuntu 26.04 arm64, `ros:rolling-ros-base@sha256:812a4b6ab03877842545398d99cf15dc25a01ff8065e8c1e117cd3750c7ca5c8`
  - source workspace: this PR at `72390b1`; `ur_client_library` `master` at `f4f8870`; `ur_msgs` `rolling-devel` at `1dbe819`; `ur_description` `rolling` at `89bbe79`
  - ros2_control-related source repositories were omitted as suggested; required Rolling binary dependencies were installed with `rosdep`

```bash
colcon build --symlink-install \
  --packages-select ur_client_library ur_msgs ur_description ur_dashboard_msgs ur_robot_driver \
  --parallel-workers 4 \
  --cmake-args -DBUILD_TESTING=ON -DUR_ROBOT_DRIVER_BUILD_INTEGRATION_TESTS=OFF

cd build/ur_robot_driver
ctest --output-on-failure -R '^test_test_urscript_interface_shutdown.py$'
```

Build result: 5 packages finished, including `ur_robot_driver` (3 min 42 sec total). Regression result: 1/1 passed in 5.83 sec.

The Rolling run exposed that `launch_testing` searches stderr by default while the URCL retry line is emitted on stdout. The regression now specifies `stream="stdout"`; no production behavior changed.

## Maintainer direction

The proposed driver-local scope was described on #838, and @urfeex confirmed that the issue still persists and invited this approach: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/838#issuecomment-5493880410

The Rolling workspace used above follows @urfeex's validation guidance: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1951#issuecomment-5524882452

---

> [!NOTE]
> **Medium Risk**
> Changes ROS 2 node lifecycle and primary-client connection timing (interfaces appear only after connect), but scope is limited to `urscript_interface` startup/shutdown—not motion, safety, or script execution paths after spin.
> 
> **Overview**
> **`urscript_interface` no longer blocks in the constructor on `PrimaryClient::start()`**, so Ctrl-C during the URCL retry wait can exit cleanly instead of hanging before `spin()`.
> 
> Startup runs on a **background thread** with mutex-tracked state. **`rclcpp::on_shutdown`** and the destructor call **`stop_primary_client_startup()`**, which invokes **`PrimaryClient::stop()`** when a connect retry is in progress, then joins the worker. **Topic and action servers are created only after startup succeeds** via **`initialize_interfaces()`**. **`main()`** rethrows stored startup failures as fatal exits when ROS is still running, and treats shutdown during startup as **exit 0**.
> 
> A new **launch test** (`test_urscript_interface_shutdown.py`) reserves loopback port 30001, waits for the retry log on stdout, sends **SIGINT**, and asserts shutdown with exit code 0. **CMake** wires the test with a 30s timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72390b1152b3726e0575a9b9e9fb02cfce6a8b3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ur_robot_driver/src/urscript_interface.cpp
+++ b/ur_robot_driver/src/urscript_interface.cpp
@@ -37,7 +37,11 @@
 
 #include <ur_client_library/primary/primary_client.h>
 
+#include <atomic>
 #include <chrono>
+#include <exception>
+#include <memory>
+#include <mutex>
 #include <thread>
 
 #include <ur_msgs/action/send_script.hpp>
@@ -58,9 +62,58 @@ public:
 
     primary_client_ =
         std::make_unique<urcl::primary_interface::PrimaryClient>(this->get_parameter("robot_ip").as_string(), notif_);
+  }
 
-    primary_client_->start(10, std::chrono::seconds(10));
+  void start_primary_client_async()
+  {
+    std::lock_guard<std::mutex> lock(startup_mutex_);
+    startup_thread_ = std::thread([this]() {
+      {
+        std::lock_guard<std::mutex> lock(startup_mutex_);
+        if (startup_stop_requested_) {
+          return;
+        }
+        startup_in_progress_ = true;
+      }
 
+      try {
+        primary_client_->start(10, std::chrono::seconds(10));
+      } catch (...) {
+        std::lock_guard<std::mutex> lock(startup_mutex_);
+        startup_exception_ = std::current_exception();
+      }
+
+      std::lock_guard<std::mutex> lock(startup_mutex_);
+      startup_in_progress_ = false;
+    });
+  }
+
+  void stop_primary_client_startup()
+  {
+    bool stop_primary_client = false;
+    {
+      std::lock_guard<std::mutex> lock(startup_mutex_);
+      startup_stop_requested_ = true;
+      stop_primary_client = startup_in_progress_;
+    }
+
+    if (stop_primary_client) {
+      primary_client_->stop();
+    }
+  }
+
+  std::exception_ptr join_primary_client_startup()
+  {
+    if (startup_thread_.joinable()) {
+      startup_thread_.join();
+    }
+
+    std::lock_guard<std::mutex> lock(startup_mutex_);
+    return startup_exception_;
+  }
+
+  void initialize_interfaces()
+  {
     script_sub_ = create_subscription<std_msgs::msg::String>(
         "~/script_command", 1, std::bind(&URScriptInterface::script_callback, this, std::placeholders::_1));
 
@@ -73,6 +126,10 @@ public:
 
   ~URScriptInterface() override
   {
+    if (startup_thread_.joinable()) {
+      stop_primary_client_startup();
+      startup_thread_.join();
+    }
     if (execute_thread_.joinable()) {
       execute_thread_.join();
     }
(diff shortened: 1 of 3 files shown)


urfeex:
@shin4141 for a development env I would suggest using ROS Rolling with a workspace containing all projects as defined inside https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/Universal_Robots_ROS2_Driver.rolling.repos (You can skip the ros2_control things right now, I think) But especially the ur_client_library and ur_msgs should best be built from their HEAD branches for any incoming feature.

By default I don't review draft PRs as I don't know whether things are in a state where the author plans to change things around. If you want feedback in a draft state, please let me know explicitly.

ur_robot_driver/src/urscript_interface.cpp
```diff
@@ -58,9 +62,58 @@ class URScriptInterface : public rclcpp::Node
...
+    }
+  }
+
+  std::exception_ptr join_primary_client_startup()
+  {
+    if (startup_thread_.joinable()) {
+      startup_thread_.join();
+    }
+
+    std::lock_guard<std::mutex> lock(startup_mutex_);
```
urfeex, line 111: From my understanding, that lock is actually useless, since you're returning a pointer to the exception. After the lock releases, anyone can alter the content of that pointer again. This is ok given the structure you've created, but the lock here doesn't do anything apart from blocking until it can be locked, which is potentially creating a deadlock.

ur_robot_driver/src/urscript_interface.cpp
```diff
@@ -58,9 +62,58 @@ class URScriptInterface : public rclcpp::Node
...
+        if (startup_stop_requested_) {
+          return;
+        }
+        startup_in_progress_ = true;
+      }
+
+      try {
+        primary_client_->start(10, std::chrono::seconds(10));
+      } catch (...) {
+        std::lock_guard<std::mutex> lock(startup_mutex_);
```
urfeex, line 82: If `startup_in_progress_` was an atomic bool, the lock wouldn't be necessary.

urfeex (commented):
I think, the general idea is good, but a couple of remarks:

- There are a lot of places with a mutex, where it's potentially easy to deadlock.
- In many places mutexes wouldn't be needed when using atomic types such as `std::atomic<bool>`.
- The race condition between start and stop as copilot has pointed out would probably something to fix in URCL.

shin4141:
Thanks for the review. I investigated the mutex and start/stop concerns in more detail.

The driver-local mutex can be removed: the two state flags can be atomic, and the startup exception can be read after joining the worker. I can also separate runtime callback exceptions from startup failures.

However, the immediate stop-before-start race cannot be closed safely in the driver alone. It appears to require a lifecycle/cancellation handshake in URCL, from `PrimaryClient::start()/stop()` through the pipeline and active connection path. Otherwise, a late `start()` can erase an already completed stop request, and the driver may still wait through the connection timeout during shutdown.

Before expanding the scope, would you prefer that I prepare a separate, narrowly scoped URCL PR for this prerequisite and keep this driver PR in draft until it lands?

The proposed URCL validation boundary would cover stop-before-connect, stop-during-connect, fast-connect run/stop, intentional restart after stop, and the driver’s immediate-SIGINT path.

shin4141:
Hi @urfeex, following up on my September 4 question about the URCL prerequisite.

Would you prefer that I prepare a separate, narrowly scoped URCL PR for the start/stop cancellation handshake, or is that work already planned or underway on your side?

I’m keeping this driver PR in draft pending that direction. Although it is a draft, I’m explicitly requesting feedback on this scope question so I can coordinate the next step.

Thanks for your earlier review and guidance.

urfeex:
> Would you prefer that I prepare a separate, narrowly scoped URCL PR for the start/stop cancellation handshake, or is that work already planned or underway on your side?

Yes, a URCL PR on that would be great!

PR #1981, opened 2026-09-14
[ur_calibration] Added call to Pipeline::init() so the primary stream co…

tonybaltovski:
…nnects before Pipeline::run() is called.

`calibration_correction` never connected to the robot and so could never extract calibration data but rather it just looped forever on `Stream is invalid / Reconnect called on a socket that is not in LostConnection state`.


urfeex (approved):
Absolutely right, thanks for fixing this. We seem to lack an integration test, there.

PR #1987, opened 2026-09-15
Migrate ur_robot_driver off deprecated Handle raw-pointer API

destogl:
## Summary
- Removes the deprecated raw-pointer export_state_interfaces()/export_command_interfaces() overrides. Every interface this driver exports (joints, sensors, all 14 gpio components) is already URDF-declared in ur.ros2_control.xacro, so the framework's default URDF-driven export now handles all of it - no export_unlisted_*_interface_descriptions() override needed.
- Interface names are built once in on_init() via a new build_interface_names(), then looked up by name via set_state()/get_command()/set_command() in read()/write() and the various async callbacks/threads, instead of aliasing member memory through raw pointers removed upstream in ros-controls/ros2_control#3610.
- Adds test_hardware_interface_exports.cpp, confirming the new export path produces the exact same interface names the old raw-pointer export used to produce.

## Test plan
- ament_add_gmock(test_hardware_interface_exports) passes: both ExportsExpectedStateInterfaces and ExportsExpectedCommandInterfaces.
- CI / maintainer review against a real robot or URSim would still be valuable, since this touches every read()/write() cycle and the async IO thread.


urfeex:
Thank you for addressing this @destogl. Would you like to continue this or should we take over driving that forward. I would like to use that opportunity to clean up quite a bit of tech debt that we have around interface handling.

